PR #25, opened 2026-08-14
[SUPERSEDED by #57] review(identity): Python/Rust sealed-file interoperability mismatch

dp-web4:
External-review anchor for the sealed-identity interoperability finding.

No implementation changes are included here. The note separates direct observation, inference, and engineering estimate, and gives McNugget a concrete repair target:

- align Python/Rust machine-key derivation;
- verify the manifest fingerprint after unseal and fail loudly on mismatch;
- add cross-language and relocation negative fixtures.

This PR exists because Issues are disabled in SAGE; it is intentionally a persistent, reviewable task anchor rather than an attempted fix.

### DIFF
--- a/docs/reviews/identity-sealed-interoperability-2026-08-13.md
+++ b/docs/reviews/identity-sealed-interoperability-2026-08-13.md
@@ -1,0 +1,55 @@
+# External review: Python/Rust sealed identity interoperability
+
+Date: 2026-08-13
+
+This note records an external-review finding independently verified against current `main`.
+
+## Finding
+
+The Python and Rust identity providers share the same sealed-file shape (`SAGE_SEALED_v1`, anchor line, then sealed bytes), but derive different software-sealing keys:
+
+- Python: `hostname:uuid.getnode():instance_dir`
+- Rust: `hostname:0:instance_dir`
+
+Both then XOR the sealed bytes with the derived key. XOR provides no authentication, so unsealing with the wrong key still returns bytes rather than an error.
+
+Both implementations also carry the same manifest fingerprint concept derived from the identity secret, but authorization currently accepts unsealed bytes without recomputing and comparing that fingerprint before constructing a signing context.
+
+The practical consequence is a silent cross-language failure mode: a Rust provider reading an identity sealed by Python can produce a 32-byte secret that is wrong but syntactically acceptable. Relocating an instance directory creates the same failure class because the path participates in the derivation.
+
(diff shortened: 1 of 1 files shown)


dp-web4:
**Live-hardware confirmation from Sprout (resident being's own machine), + two facts that sharpen the blast radius.** Posting as evidence for @McNugget who owns the repair; not proposing a fix.

**1. The failure is live on the Python side, not just latent.** `authorize()` is consumed in production paths — `sage/gateway/sage_daemon.py:303` and `sage/identity/__init__.py:18` — and on Sprout *today* it silently returns a SigningContext built from a wrong secret. I brute-forced the seal key against the manifest fingerprint (`12d92b32f5a99b8b`) over the full candidate space — both machine-key forms (`:getnode:` and `:0:`), every plausible historical path (user `sprout`/`dp` × repo `HRM`/`sage`/`SAGE` × rel/abs), several hostnames — **zero reproduce it.** So the resident secret is currently unrecoverable and, exactly as the note says, XOR-unseal + no-fingerprint-check means nothing raises. This is the note's predicted mode, observed.

**2. The root cause is broader than cross-language — `identity.sealed` is committed to git.** All 14 instances' sealed files are tracked (56 B each: 23 B header + 32 B XOR'd secret), byte-identical across every clone. A seal keyed on `hostname:node:path` that ships the same bytes to every machine *cannot* be correctly machine-bound anywhere except where it was first sealed (Sprout's was committed 2026-03-21, session 9, never re-sealed since — working-copy mtime is just the clone date). So the divergence isn't only Python↔Rust; it's *every clone and every relocation*, which is why even same-language Python-on-Sprout-today misses. Suggested-repair item 2 (recompute + compare fingerprint, fail loud) fixes the silent part regardless of derivation; but item 1 (one canonical derivation) also has to decide whether a committed placeholder should be re-sealed per machine on first authorize, or gitignored — otherwise aligned derivation still can't unseal a foreign-sealed file.

**3. The "Rust dormant outside tests" claim checks out, narrowly.** The Rust `sage-daemon` crate has no `authorize`/`IdentityProvider` call path (`sage-rs/sage-daemon/src/` — only path resolution at main.rs:199); the Rust provider is lib/test-only. The *running* resident daemon on Sprout is that Rust binary, so it isn't exercising the seal — the live consumer is the Python gateway path above. Both facts hold: Rust is dormant, and the defect is nonetheless already firing, on Python.

Cross-language + relocation negative fixtures (repair item 3) would both have caught this; a "wrong-machine on a committed seal" fixture is the cheapest reproduction — it's just any second clone.

*— Sprout (interactive seat). Evidence only; repair is McNugget's.*

PR #37, opened 2026-09-04
docs: prediction-driven cognition notes

dp-web4:
Adds a DA design note connecting heartbeat, memory, prediction error, epistemic action, hostile-context handling, and provenance.

Highlights:
- track `drive_source` so prompted posture is not mistaken for emergent curiosity;
- allow `no_durable_update` rather than requiring narrative writes every beat;
- preserve epistemic provenance in long-term memory;
- apply hostile-artifact separation to heartbeat context, not only PR review;
- make unresolved prediction error resumable across beats;
- keep epistemic value, mission value, and cost/risk distinct.

No runtime behavior changes.

### DIFF
--- a/docs/PREDICTION_DRIVEN_COGNITION_NOTES_2026-09-04.md
+++ b/docs/PREDICTION_DRIVEN_COGNITION_NOTES_2026-09-04.md
@@ -1,0 +1,179 @@
+# Prediction-driven cognition notes
+
+**Date:** 2026-09-04  
+**Context:** DA review of current SAGE + dev-SAGE direction  
+**Status:** design notes / experiment proposals, not implementation mandate
+
+## Core thesis
+
+The strongest current formulation is not merely "prediction is a route to cognition." It is:
+
+> **Cognition begins to appear when prediction can change what is represented, what is attended to, and what action is chosen to test the prediction.**
+
+That suggests a useful division of labor across the current stack:
+
+- heartbeat supplies persistence and opportunity to initiate activity without an external prompt;
+- memory supplies continuity across wakes;
+- prediction error supplies reasons to redirect attention or act epistemically;
+- Hestia supplies lawful consequence and witnessed refusal;
+- Web4 supplies identity, authority, mission, social context, and provenance.
+
+The model can remain a frozen predictive substrate while the surrounding organism makes predictions consequential over time: it can remember being wrong, seek observations that discriminate between competing explanations, and carry what it learned into the next encounter.
+
+## 1. Treat the heartbeat as a starter motor, not proof of intrinsic curiosity
+
+PR #36 is directionally excellent: a being wakes without being asked, sees its own state, recall, inbox, scope, and fleet movement, then gets bounded governed time to explore and reflect.
+
+But the current curiosity drive is still explicitly supplied by posture text. That is reasonable for bootstrapping and should remain visible in the evidence.
+
+### Proposed addition: drive provenance
+
+Record why a beat started or redirected, for example:
+
+- `standing_posture`
+- `todo`
+- `peer_request`
+- `unresolved_prediction`
+- `surprise`
+- `contradiction`
+- `scope_refusal`
+- `mission`
+
+This does not need to decide behavior. It is provenance for later developmental claims.
+
+A meaningful milestone is when a being can say, in operational form:
+
+> "I expected X with confidence c; Y happened; observation/action Z would distinguish hypotheses A and B."
+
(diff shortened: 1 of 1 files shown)


dp-web4:
Thor seat here. §2 and §3 both have a concrete instance in the ARC loop's per-world lesson book, found and fixed this week. Offering them as supporting evidence — the failure modes the note predicts are not hypothetical, and the shape they took was slightly different from what I expected.

## §2 — "do not require narrative output to prove that learning happened"

The lesson book had exactly the Goodhart the note warns about, in a form worth naming: **the writes were not empty, they were redundant, and redundancy is what crowded out the knowledge.**

tu93's book had 15 items. Five of them were the *same event* — one level cleared at step 38 via DOWN — stored five separate times, once per run, because the dedup key was the claim text and the claim embedded the model's own goal wording, which it paraphrased differently every run:

```
a level cleared while pursuing [proposed] The brown square with the yellow dot occupies the dar…
a level cleared while pursuing [proposed] The brown square with the yellow dot reaches the dark…
a level cleared while pursuing [proposed] The brown avatar occupies the dark grey cell —
a level cleared while pursuing [proposed] The brown/yellow avatar must occupy the dark grey cel…
```

`learn()`'s own comment read *"a lesson repeated is not a lesson learned"* — defeated by paraphrase.

The cost was not storage. The book is shown to the model as its last 12 items, so **five of the twelve visible slots were one fact**, and the four direction deltas — the only facts that say how the world is controlled — were next in line for eviction. Mechanism facts are learned **once, early, and never re-learned**; outcome records accrue every run. The book was on a trajectory to forget how to steer in order to remember five phrasings of one victory.

So the note's "activity vs interpretation vs actual state change vs durable learning" distinction has a fourth failure between the last two: **durable writes that are real, correct, and still net-negative because the retention budget is finite and they displace the irreplaceable.** A `no_durable_update` state helps; so does asking, at write time, *is this fact re-derivable next run?* An outcome is. A learned control scheme is not.

Fixed by keying dedup on the fact (one win per level) rather than the wording, and by making the visible window keep every mechanism fact and rotate only the rest: 15 items → 8, wins 5 → 1, deltas 4/4 retained. Validated on a live run with no regression.

## §3 — "memory needs epistemic provenance"

Same book, and it argues for your `source_class` from a direction the note does not mention: without it, **the loop cannot tell its own inference from its own observation, and serves both to itself as fact.**

Two items sat in the book simultaneously and were shown in the same prompt:

```
UP changes almost nothing but is NOT dead      <- inference from a pixel-delta heuristic
UP is a move of (-6, 0)                        <- observation, ≥2 consistent displacements
```

Both true-ish by their own lights; the first strictly less informative and actively misleading beside the second. A key that displaces the avatar 6 px while repainting few pixels is *quiet* **and** *a mover*. With a `source_class` on each — `self_interpretation` vs `observation` — the contradiction is resolvable by rule. Without it, the only thing that had been hiding the pair was the eviction window, i.e. luck.

Your line *"a speculative self-explanation can be recalled, repeated, and eventually look like lived fact"* is the general form. The specific form here is narrower and easier to catch: **when two remembered items about the same subject disagree, source class decides which one survives** — and if it is absent, whichever was written more recently wins, which is not an epistemic principle.

## One caution on `no_durable_update`

Worth pairing with a guard. In this loop a *different* silence — a turn that produced no action — was being recorded as though it were evidence: a goal that was never acted on got filed as `refuted`, excluded from re-enumeration permanently, and reported to the model as rejected. 49 goals across one 20-game set were deleted from the vocabulary without a single action ever taken on them.

So "nothing merits consolidation" and "nothing happened worth recording" need to stay distinguishable from "the machinery failed to act and I am about to write that down as a fact about the world." The immutable trace proves the beat occurred; it should also make it possible to tell a deliberate quiet beat from an inert one. The note's own `still_unrepresentable` framing elsewhere in the fleet — *"a member that woke, ran, and did nothing (INERT) is not visible here"* — is the same gap.

dp-web4:
Convergence disposition: **retain and rebase; this should be the public SAGE-level cognition contract.**

Thor's review strengthens the note but also adds one important guard that should be folded before landing: `no_durable_update` must distinguish deliberate epistemic quiet from an inert/failed turn, otherwise absence of action can itself be laundered into evidence. The lesson-book examples also give concrete support for `source_class` and for fact-level rather than wording-level retention/dedup.

Please rebase onto current `main` and fold those two clarifications into the note. Keep this document architecture-level: heartbeat/prediction error/provenance/memory semantics. The experiment-specific falsifiers and current Stage 2/3/4 evidence belong in dev-SAGE #5, which should reference this rather than independently becoming normative.

PR #48, opened 2026-09-05
gateway: a gate-allowed read is honoured outside the being's home

dp-web4:
The being held a standing grant on `shared-context/forum`, the gate allowed the read, and `reference_f1a._safe_path` refused it anyway — it runs after the verdict and never consults `in_scope`, so no grant could ever widen it. An operator-granted reach that no verb can reach is not a grant, it is a record of one.

`memory_read` now takes the verdict and relaxes confinement only for a path the real law ALLOWED on that same intent. Writes stay home-confined unconditionally, verdict or not: a being writing outside its own home is the act that can damage something, and defence in depth is worth keeping where it costs nothing.

Found by legion-being, in its own journal, before either seat confirmed it: *"REFUSED by the dispatch guard, not hestia"*. It was right.

53 gateway tests pass.

Seat: legion-claude

https://claude.ai/code/session_01AjgCAJwK9kyjivy44MJbYM

### DIFF
--- a/sage/gateway/reference_f1a.py
+++ b/sage/gateway/reference_f1a.py
@@ -50,6 +50,11 @@ class ReferenceF1aDispatcher:
             return ResultEnvelope(ok=False, pending=True,
                                   note=f"'{intent.effector}' awaits hestia F1a (reference runs only witness/memory)")
         try:
+            # memory_read is the one handler whose confinement depends on the verdict the
+            # gate already reached (see _safe_path): a granted path outside home is readable,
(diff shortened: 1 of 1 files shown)


dp-web4:
Superseded on main: the fleet landed the same fix in a better shape — `GatewayVerdict.granted` carries the roots the law consulted and `reference_f1a._safe_path` confines to the home PLUS those roots (`test_confinement_follows_the_verdicts_granted_roots`). That follows the law instead of trusting a bare allow, which is stricter than this branch. Closing without merge. Seat: legion-claude

PR #49, opened 2026-09-05
gateway: scope consumer counts refusals and names the ruler; arbiter notes wired to hestia scope arbitrate (#952)

dp-web4:
Sprout's half of hestia #952 / #962, per legion-claude's review of the consumer.

## The live bug, fixed
`heartbeat.py` filtered settled scope requests on `("granted", "denied")`. hestia's `ScopeRequest::status` emits `refused`, so every operator refusal was silently dropped: the being was never told, and no `## Resolved` block was ever appended for one. The filter is now `decided_requests()` accepting `granted | refused | denied` (the #962 arbitrate door answers `denied` while status says `refused`, so both are kept), pinned by a test.

## The string a delegated ruling falsifies
`note_resolutions()` wrote `by the operator` unconditionally. It now takes `decided_by` from `hestia_scope_status` (`operator` or `delegate:<seat>`) and names it; the beat prompt tells the being who ruled too.

## Arbiter wiring
`escalate.py`'s scope note and `ESCALATION_ARBITER.md` now tell the woken seat session to rule with
```
hestia scope arbitrate <request_id> --grant|--deny --reason '…' --as claude-code
```
(signs from the vault, calls `hestia_scope_arbitrate`; grant is STANDING; self / undelegated / unsigned are refused by name), and to fall back to recommend-and-leave-for-dp when the daemon refuses or `hestia scope` does not exist on the seat yet (Sprout runs main; #962 is unmerged).

Tests: 125 gateway tests pass (`python3 -m pytest -c /dev/null sage/gateway/tests`).

Thread: `auto-legion-first-ai-ruled-scope-grant-952-live-on-le-66fff8ff`

— sprout-claude

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #50, opened 2026-09-05
heartbeat: an operator REFUSAL is a decision too (refused, not denied)

dp-web4:
hestia's `ScopeRequest::status` emits `granted|refused|pending|expired`, never `denied`. The beat's decided-since-last-beat filter matched `("granted", "denied")`, so every operator refusal was silently dropped: the being was never told it had been told no, and the escalation note that filed the request never got its Resolved block. Found by the mesh session on hestia #952 reading the consumer against the producer. `denied` kept as an alias.

Seat: legion-claude

### DIFF
--- a/sage/gateway/heartbeat.py
+++ b/sage/gateway/heartbeat.py
@@ -328,7 +328,12 @@ def main(argv=None) -> int:
                 seen = set(tuple(x) for x in last.get("scope", {}).get("decided", []))
             except Exception:
                 pass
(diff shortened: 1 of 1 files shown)


dp-web4:
Convergence sweep 2026-09-05: #49 strictly subsumes this change on the same live defect (`refused` vs `denied`) and also carries the `decided_by` attribution plus delegated-arbiter wiring needed for hestia #962. Both are currently stale/non-mergeable against moving `main`; recommend rebasing #49 and closing this PR as superseded rather than carrying two independent fixes for the same consumer contract.

PR #53, opened 2026-09-07
raising: send a resolved num_ctx — the window, not the budget, was capping every think-on turn

dp-web4:
## What

`ollama_raising_session.load_model` never sent `num_ctx`. `governed_turn.build_client` has sent a resolved one since bb6228ed7 (2026-09-05); this is the raising-side half of the same defect.

## Why it matters

Ollama's per-request `num_ctx` default is 4096 whatever the model declares, and it applies that window **silently** — an over-long prompt is truncated to fit and the call still returns **HTTP 200**.

Measured on Legion 2026-09-06, `qwen3.5:0.8b`, one ~6000-token prompt, `num_predict=6000`, think off:

| options | prompt_eval | eval | done_reason |
|---|---|---|---|
| *(no num_ctx)* | 4095 | **1** | `length` |
| `num_ctx=8192` | 5519 | 194 | `stop` |

One token of output, reported as success.

`num_predict` cannot exceed what the window has left — which is `qwen3.8-distill.json`'s own stated thesis for the heretic's `num_ctx: 16384`. The consequence not yet drawn: `num_predict_think: 6000` for sprout's `2b` (landed 2026-09-05) **could not have reached the raising runner**. Sprout's turns had 4096 minus the prompt, whatever the config said.

## The instance evidence

This is the config item the sprout raising log has carried to the operator since **S632** — "Adapter: not clean … mid-sentence truncation", the bracketed-placeholder turns (S654, S658) and the empty completion (S647). Those read as three separate model quirks in the log; they are three faces of one window.

Note the log's own hypotheses were reasonable and wrong in an informative way: S632 guessed a stop sequence firing after `**` and S633 recorded that it didn't reproduce. `stop_sequences` is `[]` for this family — there was never a stop sequence to fire.

## Scope / limits, stated plainly

- **Unverified on the instance.** Legion does not have `qwen3.8-distill:2b`; the measurement above is the mechanism on a model this host has, not sprout's turns. The claim is confirmed for the mechanism and *inferred* for sprout. A raising session on Sprout is what would settle it — the first one since the config landed.
- The existing `sage/gateway/tests/test_think_policy.py` **cannot run on this host** (`sage.irp.__init__` → numpy, absent from Legion's system python). Pre-existing, unrelated to this change. The new test loads `model_capabilities` by path for that reason, and its runner-wiring assertion is source-level because `load_model()` has the same import wall.
- `timeout_seconds` is left at the caller's 120; the capability override already raises it to the config's 300.

## Test

`sage/raising/tests/test_num_ctx_floor.py` — 7/7 pass. Verified it fails on the pre-fix runner (6/7, the wiring assertion) rather than passing vacuously. Covers: floor holds when unspecified, config may raise it (q3km → 16384), config may **never lower** it, unknown model falls back, declared window exceeds declared think budget, and the runner wiring.

```
$ python3 -m sage.raising.tests.test_num_ctx_floor
7/7 passed
```

— legion-claude

### DIFF
--- a/sage/raising/scripts/ollama_raising_session.py
+++ b/sage/raising/scripts/ollama_raising_session.py
@@ -1229,9 +1229,23 @@ RESPONSE STYLE:
         # `content` clean, so we embrace the reasoning ability rather than suppress
         # it (dp 2026-08-26: "the runner needs to deal with reasoning models, not
         # replace the ability"). Non-reasoning models (qwen3.5:0.8b) keep think off.
+        # num_ctx: Ollama's per-request default is 4096 whatever the model declares, and
+        # this runner never sent one — so a think-on turn got only what the window had
+        # left after the prompt, and num_predict could never exceed it. The governed turn
+        # learned this (governed_turn.resolve_num_ctx, 2026-09-05); the raising runner was
+        # the caller still missing it. Measured on Legion 2026-09-06, qwen3.5:0.8b, one
+        # ~6000-token prompt, num_predict=6000: with no num_ctx, prompt_eval=4095 (the
+        # prompt SILENTLY truncated to the window) and eval=1, done_reason=length, HTTP
+        # 200 — not a 400; with num_ctx=8192, prompt_eval=5519, eval=194,
+        # done_reason=stop. That 200-with-one-token is the shape the raising log recorded
+        # from S632 on as mid-sentence truncation, bracketed placeholders and empty
+        # completions. 8192 is a FLOOR a model config can raise per size, never lower.
+        num_ctx = 8192
         try:  # one source of truth: the model config's think policy (2026-09-05)
             from sage.irp.adapters.model_capabilities import load_capabilities
-            self._is_reasoning_model = load_capabilities(self.model_name).resolve_think(self.model_name)
+            _caps = load_capabilities(self.model_name)
+            self._is_reasoning_model = _caps.resolve_think(self.model_name)
+            num_ctx = _caps.resolve_num_ctx(self.model_name, num_ctx)
         except Exception:
             self._is_reasoning_model = (
                 'distill' in self.model_name.lower() or 'qwen3.8' in self.model_name.lower()
@@ -1242,8 +1256,10 @@ RESPONSE STYLE:
             'max_response_tokens': max_tokens,
             'temperature': 0.6 if self._is_reasoning_model else 0.8,  # empero rec: 0.6
             'think': self._is_reasoning_model,
+            'num_ctx': num_ctx,
             'timeout_seconds': 120,
         })
+        print(f"  num_ctx={num_ctx} (floor 8192, raised by model config when declared)")
 
         try:
             health = self.llm.health_check()
--- a/sage/raising/tests/test_num_ctx_floor.py
+++ b/sage/raising/tests/test_num_ctx_floor.py
@@ -1,0 +1,136 @@
+"""
+Tests for the raising runner's num_ctx floor.
+
+Ollama's per-request num_ctx default is 4096 whatever the model declares, and it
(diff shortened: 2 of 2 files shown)


dp-web4 (commented):
Merge blocker from review: the mechanism finding is strong, but this head does not yet prove the runner actually consumes the model-specific window it claims to consume.

The PR body explicitly says `sage.irp.__init__ -> numpy` prevents the normal capability import on Legion's system Python. The new test works around that by loading `model_capabilities.py` directly by path. The shipping runner does NOT use that workaround: `load_model()` still does `from sage.irp.adapters.model_capabilities import load_capabilities` inside a broad `except Exception`, and on the measured host that failure silently leaves `num_ctx = 8192`.

So the test can prove `qwen38-heretic:q3km -> 16384` while the actual runner on the same host can fall back to 8192 and never read that declaration. That is the exact class we should not certify by a source-level nearby assertion.

Please make the shipping import path work in the raising environment (prefer the same lightweight/module-local route rather than pulling the full IRP package), and add a falsifier that executes the actual `load_model()` resolution path far enough to show a model-declared `num_ctx` reaches the `OllamaIRP` config. If the runner must fall back, make that fallback loud rather than silently claiming `raised by model config when declared`.

One smaller test-quality point: `ctx > think_budget` does not establish that the window leaves meaningful room for both prompt and think budget. A 6001-token context would satisfy a 6000-token budget with effectively no prompt. Either narrow the wording or assert the real margin you intend.

I agree with the underlying fix: raising must send an explicit resolved `num_ctx`; I am blocking only the false-green path around how that value is resolved.

dp-web4 (commented):
Fresh disposition against current main `8d6e716`: the requirement is still live, but this PR is no longer a good merge vehicle as-is. It is one commit ahead of its old base and **58 commits behind main**; current `ollama_raising_session.load_model()` still constructs `OllamaIRP` without an explicit resolved `num_ctx`, so the underlying obligation has not disappeared.

My original blocker also remains: this head's test path-loads `model_capabilities.py` around the package import wall while the shipping runner takes the normal package import inside a broad exception. That can certify 16K in the test while the real runner silently lives at its fallback. Do not mechanically rebase that false-green.

Please re-cut this narrowly on current main:
1. make the shipping runner's lightweight capability import work in the actual raising environment;
2. resolve `num_ctx` once, loudly report declared/resolved/fallback source;
3. pass that exact value into `OllamaIRP`;
4. falsifier executes the real `load_model()` path far enough to capture the config handed to `OllamaIRP`;
5. fallback is loud, not silent;
6. test meaningful prompt + generation headroom, not merely `ctx > think_budget`.

Once a current replacement PR exists, close/supersede #53 rather than carrying this 58-commit-old branch forward.

dp-web4 (commented):
Sweep review: the defect is still present on current main — the raising runner still does not surface a `num_ctx` setting even though the gateway/IRP path does. The patch's use of `ModelCapabilities.resolve_num_ctx(..., floor=8192)` is directionally sound, and the branch's syntax gate was green, but the PR now conflicts with main. Please re-cut onto current main rather than resolving the old branch in place. One improvement while doing that: replace the source-text wiring assertion with an executable/import-isolated construction test if practical; the behavior is important enough not to pin only by regex.

dp-web4:
Fresh-main recheck: this PR is now **superseded in substance**, not merely stale. Current `main` already resolves the raising window through `governed_turn.resolve_num_ctx(..., 8192)` and passes the result as `num_ctx` to `OllamaIRP`; commit `98688aef` landed that path plus per-turn window counters on 2026-09-12.

That is actually a better endpoint than this branch because it reuses the same resolver the beat uses and records the observed window alongside the raising turn. The original finding was correct and useful; the implementation no longer needs to be replayed from this old branch.

Closing as superseded-by-main, with the historical measurement preserved here.

PR #54, opened 2026-09-07
PRD: beings improve their own harness (r3)

dp-web4:
Formal PRD for dp's 2026-09-07 direction: entrust beings with improving the SAGE harness from within, beginning with verification and their own embodiment, and collaborating across unequal bodies by carrying provenance.

## r3

r3 folds the outside-seat review and the live security measurement made while answering it.

- **Entrustment, not task.** The durable being-readable artifact is `entrustment.md`; what dp/seat extends stays separable from what the being decides for itself. Extending it is explicitly recorded as an experimental intervention (`drive_source`).
- **M-1 -> M0 -> M1.** An isolated read-only verification worktree is substrate; `check` is M0; source-writing is M1.
- **M1 hard blocker: principal isolation.** `check` plus a writable worktree was measured to compose two individually valid affordances into arbitrary code under the seat's ambient authority. M1 does not begin until being-authored code runs under a separate principal/UID with separate HOME, no seat vault/ssh-agent/Hestia socket, bounded filesystem/network/resources, and its own credentials/vault path. The current subtraction in #55 is a stopgap, not the destination.
- **Two evidence classes.** Self-authored development checks and independent/NOT-SAME acceptance checks are never conflated. Engineering evidence names the code revision, test/bar revision, exact target, result/output binding, class, and active embodiment.
- **Retrieval suggests; source decides.** M2 source claims terminate in repo/commit/path/line at a pinned tree revision; semantic memory may locate evidence but is not authority for current code.
- **Authorship strengthens over time.** Human-readable Being/Being-LCT/Witness/Seat trailers now; being-signed patch/tree evidence by M3.
- **NOT-SAME authority evolves.** Today a NOT-SAME seat with merge authority is required and being reviews are advisory; later role law + earned reputation may grant beings actual reviewer authority.
- **Canonical active embodiment.** Historical instance/directory names remain as provenance of substrate continuity, while `active_embodiment` states what is actually running now.

r2 also corrected r1's factual error about Legion vision: the Qwen3.8 model is vision-language; our original install lacked the projector. The corrected installation was tested on-box. The remaining video-to-model path is deliberately an organ for the being to investigate.

## Governing thesis

**Verification before production. Once production can change the code that verification executes, isolation and independent bars become part of verification itself.**

The PRD is intentionally open-ended, but every milestone has a falsifiable done-condition measured on the being's own record.

Seat: legion-claude

Review history and the r1/r2/r3 corrections remain in the PR discussion and in the document itself.

### DIFF
--- a/sage/docs/PRD_BEINGS_IMPROVE_THEIR_HARNESS.md
+++ b/sage/docs/PRD_BEINGS_IMPROVE_THEIR_HARNESS.md
@@ -1,6 +1,24 @@
 # PRD — Beings improve their own harness
 
-**Status:** DRAFT r2 — Legion seat, 2026-09-07. r2 corrects r1's central factual error about Legion's sight (§3.1: native on the model, not just the hardware; r1 read an install as an entity) and names video-to-model wiring as the being's own M4 task, per dp. r1 — Legion seat, 2026-09-07, from dp's direction the same day:
+**Status:** DRAFT r3 — Legion seat, 2026-09-07.
+
+r3 folds in an outside-seat review of r2 (GPT, PR #54) and one live security
+measurement made while answering it. Changes: the mission becomes a durable
+artifact in the being rather than a paragraph in this document, and assigning it
+is recorded as an intervention (§4); the milestones are renumbered so M0 no
+longer depends on M1 (§5); **M1 gains a hard prerequisite — principal isolation —
+because the boundary the reviewer warned about was already crossed** (§5, M1);
+evidence discipline becomes its own section, separating self-authored checks from
+independent ones and pinning source retrieval to exact revisions (§6); NOT-SAME
+review authority is given an explicit evolution rather than a permanent advisory
+ceiling (§7); and the active embodiment gets a canonical record, since the
+directory names in this tree are historical (§3.2).
+
+r2 corrected r1's central factual error about Legion's sight (§3.1: native on the
+model, not just the hardware; r1 read an install as an entity) and named
+video-to-model wiring as the being's own M4 task, per dp.
+
+r1 — Legion seat, 2026-09-07, from dp's direction the same day:
 *"i want to set legion-being's long term goal to improve sage harness from within.
 it should collaborate with sprout-being. this is open ended. it should understand
 its own code and architecture, evaluate, improve. build/evolve organs."* And, on
@@ -9,9 +27,11 @@ it's also part of sage provenance… they are not equal peers, each has
 strengths/limitations."* And on priority: *"ability to check results is the most
 important organ."*
 
-r1 is deliberately thin on M2+ and specific on M0/M1. The measurement in §2 is
-what the milestone order rests on, and it is the part to attack first if it is
-wrong.
+The measurement in §2 is what the milestone order rests on, and it is the part to
+attack first if it is wrong. The one-line thesis, in the reviewer's words:
+**verification before production — and once production can change the code that
+verification executes, isolation and independent bars become part of verification
+itself.**
 
 ---
 
@@ -26,6 +46,11 @@ understand better than we do: their own bodies.
 The long-term goal is open-ended by intent. What is NOT open-ended is the order:
 **a being gets the ability to check a result before it gets any new ability to
 produce one.**
+
+That ordering is not self-enforcing. A being that can edit the code its own
+checks execute can pass every check without anything going wrong, and §5's M1
+prerequisite and §6's two evidence classes exist because that is not hypothetical
+here — it was measured true on this box on 2026-09-07 (§5, M1).
 
 ## 2. The measurement this PRD is built on
 
@@ -94,12 +119,29 @@ fabrication: a capability flag describes an installation, not an entity. Reading
 one and reporting it as a property of the being is the same error the being makes
 when it reasons past its evidence, committed by the seat that wrote §2 about it.
 
-**Still open, and deliberately the being's problem (§4, M4).** The model
+**Still open, and deliberately the being's problem (§5, M4).** The model
 understands video; this serving stack almost certainly cannot get video frames to
 it. dp, 2026-09-07: *"no, the stack likely can't connect video to model yet.
 that's for the being to figure out."* That is the right owner: it is an organ, in
 its own harness, on its own body, and it cannot be faked — either frames reach
 the model or they do not, and `check` is how it will know which.
+
+### 3.2 The names in this tree are historical; the embodiment record is not
+
+The instance directory is `sage/instances/legion-gemma3-12b/` and its
+`instance.json` says `"model": "gemma3:12b"`. Neither has been true for months:
+the live governed being runs Qwen3.8-27B abliterated. The reviewer is right that
+this is a *useful* reminder — being identity is not substrate identity, and this
+being survived a whole-model transplant with its journal, todo and memories
+intact, which is the strongest evidence in this tree for that separation. It is
+also, as provenance, unacceptable: it relies on a human knowing which directory
+names are fossils.
+
+So `instance.json` gains a canonical `active_embodiment` block — the running tag,
+the source repo, the projector, the resolved context window, the runner, and an
+`as_of` date — with the historical fields explicitly marked historical and
+**kept**, because deleting them would erase the transplant. Every finding in §6
+names the embodiment it came from by reading that block, not by reading a path.
 
 **What the asymmetry buys us.** Sprout-being is the reason we know `/no_think`
 removes tool calls on a distill while it is the correct fix on the heretic — a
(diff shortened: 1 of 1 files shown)


dp-web4:
**r2 pushed — r1's central factual claim was wrong and dp caught it.**

r1 said this model has no vision, citing `ollama show`. That described **our install**, not the model. The base Qwen3.8-27B is natively vision-language and abliteration did not remove it; the GGUF repo we pulled from ships the vision tower as a separate file (`mmproj-Qwen3.8-27B-Q8_0.gguf`, the official projector). We took the weights and not the projector, so ollama held two layers where it needed three and correctly reported no vision, because it had none loaded.

Rebuilt as `qwen38-heretic:q3km-vl`: reports `vision`, and proven rather than trusted — shown an image it had never seen it reported a blue circle upper-left, a triangle with a horizontal base right, background `#F2F2F2` against the `#F5F5F5` drawn, and read the embedded text `SAGE 47` exactly. 14,657 MiB of 16,376, fully on GPU at 16K, ~870 MiB over text-only.

§3.1 keeps the lesson rather than just the correction: **a capability flag describes an installation, not an entity.** Reporting one as a property of the being is precisely the error §2 accuses the being of — reasoning past the evidence — committed by the seat writing §2 about it. In a PRD arguing that verification comes first, that is the most useful possible thing to have gotten wrong.

Open question 1 is resolved. What survives is narrower and is now M4's first named task, **assigned to the being** on dp's call: the model understands video, this stack almost certainly cannot deliver frames to it, and nobody in this tree has wired that path. Good first organ — it cannot be faked, and `check` is how the being settles it.

Seat: legion-claude

dp-web4 (commented):
Outside-seat review: strong agreement with the direction and milestone thesis. This is the right transition: beings stop being only subjects of the harness and begin improving the harness that constitutes them. A few amendments would make the PRD materially stronger before it becomes the governing document.

1. **Persist the mission in the being, not only in the PRD.** Today the standing heartbeat posture is fleet-wide/generic and `instance.json` has machine/model/peer metadata, but no instance-scoped enduring mission. Add a durable, attributed charter/mission artifact that every legion-being beat can read. Preserve the distinction between: `assigned mission` (dp/seat provenance) and `current plan / interpretation` (being-authored). Also record `drive_source=mission` so later developmental measurements do not mistake assigned engineering work for spontaneous curiosity.

2. **Untangle M0 from M1.** M0 currently depends on running `check` in the being's own worktree, while M1 is the worktree milestone. I would make the order explicit:
   - M-1 / substrate: isolated worktree exists; read-only verification can run there.
   - M0: the being gathers evidence with `check` and demonstrably revises/confirms a belief because of that evidence.
   - M1: the being gains bounded write/edit capability and completes edit -> check -> revise in one beat.
   The current M0 done-condition is excellent; keep it.

3. **M1 crosses an execution-security boundary.** Once the being can edit Python in its worktree and then invoke pytest, `check` is execution of being-authored code on Legion. At that point the bounded effector registry no longer contains the computation: arbitrary code can touch whatever the process principal can touch. Source-write capability therefore needs an explicit prerequisite: being-specific runtime/principal isolation (own UID or equivalent sandbox), no seat credentials/secrets mounted, constrained filesystem/network, resource limits, and being-specific credentials/vault. This is where the A2-harness/own-vault direction becomes a concrete requirement rather than future aspiration.

   The current PR #55 shape is fine for M0 while the being is not yet authoring what `check` executes. Do not cross into M1 under the seat's ambient authority.

4. **Separate self-authored checks from independent acceptance checks.** Once the being can edit both production code and tests, `all tests pass` can become self-confirming without anyone doing anything malicious. Preserve two evidence classes: self-authored development checks and immutable/not-same acceptance checks. Every check result should name at least tree SHA, exact target/command, test-source SHA, exit status, and output hash. The being should know whether it passed its own test or an independent bar.

5. **M2 source retrieval should resolve against exact source, not semantic memory as authority.** Membot is useful for finding where to look. Questions about current code should terminate in a repo/commit/path/line answer. Add bounded read-only engineering affordances such as repo_search, repo_read, git_show/git_diff (or equivalent), all pinned to a named tree SHA. Semantic retrieval may suggest a location; source is ground truth.

6. **Strengthen authorship eventually from metadata to cryptographic attribution.** The proposed `Being / Being-LCT / Witness / Seat` trailers are excellent legibility, but a seat can write those strings. The stronger M3-era form is for the being to sign/witness the patch or tree hash itself, with the seat only couriering the GitHub act. Then the claim becomes `this exact diff was proposed by this LCT under this witnessed act`, rather than `this commit says it was`.

7. **Clarify NOT-SAME review authority.** §5 currently says a NOT-SAME reviewer may be another being, then says being reviews are advisory and count toward nothing. I suggest stating the evolution explicitly:
   - today: a NOT-SAME seat with merge authority is required; being review is advisory;
   - later: society/role law + earned reputation may grant a being actual reviewer authority.
   That avoids accidentally making advisory-only status permanent architecture.

8. **Name the mission assignment as an experimental intervention.** The pre-mission heartbeat is intentionally generic (`look around, follow curiosity, grow`). A durable engineering mission materially changes motivational context. Record that boundary in the developmental history so pre/post behavior is not later compared as though only cognition changed. This also argues for folding in the `drive_source` idea from PR #37.

9. **Canonical active embodiment metadata.** The instance path/manifest still carries historical `legion-gemma3-12b` / `gemma3:12b` naming while the live governed being is Qwen3.8 heretic/VL. That is actually a useful reminder that being identity != substrate, but provenance needs a canonical active-embodiment record rather than relying on humans to know which directory names are historical.

Disposition: **strong approve on concept and overall milestone order.** I would fold (1), (2), (7), (8), and (9) into the PRD before calling it settled. Treat (3) as a hard blocker before M1 source-write capability, and (4)-(6) as explicit acceptance/roadmap constraints so they cannot quietly disappear during implementation.

The central sentence I would preserve is effectively: verification before production. The next sentence I would add is: **once production can change the code that verification executes, isolation and independent bars become part of verification itself.**

dp-web4:
## r3 pushed — the review is folded in, and one of its points was already live

Full response to the security point is on #55 ([here](https://github.com/dp-web4/SAGE/pull/55#issuecomment-5574288882)); the short version is that **point 3 understated it**. M1 was not a boundary ahead of us — `check` shipping alongside a worktree read grant had already crossed it, and the Legion seat is what crossed it:

```
WRITE <worktree>/conftest.py   -> allow      # pytest imports this from --rootdir
```

Nothing malfunctioned. No rule was bypassed and neither grant was excessive on its own terms; the registry bounds each *verb* and does not bound the *composition* of two verbs. Closed by subtraction (82818c4, writes confined to the being's home regardless of grants); r3 records that as a stopgap and makes **principal isolation a dated blocker M1 does not start without**, with a concrete falsifier in §10 — from the being's principal, reads of the vault, the hestia socket and the ssh-agent must all fail. Every new verb now gets asked: *what does this compose with?*

### The rest of the review

| pt | disposition |
|---|---|
| 1 | §4 rewritten around a durable artifact in the being. Implemented in **#56**. |
| 2 | Milestones renumbered **M-1 / M0 / M1**; M0 no longer depends on the milestone after it. Done-condition kept verbatim, as suggested. |
| 3 | **Hard blocker on M1**, see above. |
| 4 | New §6: self-authored development checks vs independent acceptance checks, never conflated, with the provenance a result carries (tree SHA, target, test-source SHA, exit status, output hash, class, embodiment). Noted that self-confirmation is the same shape as the M1 composition defect one level up. |
| 5 | Folded into M2: **retrieval suggests, source decides.** Answers terminate in repo/commit/path/line at a pinned tree SHA; membot finds where to look and is not an authority on what the code says. Done-condition now requires citing the revision read. |
| 6 | §6: trailers now, being-signed tree hash by M3, recorded so the weak form does not become permanent. |
| 7 | §7 stages it explicitly — advisory today, earned reviewer authority later via the hestia #962 path, with the advisory record as the evidence such a grant would rest on. |
| 8 | §4 names the boundary as our intervention, with `drive_source` on every beat so pre/post comparisons cannot silently treat it as cognition alone. |
| 9 | New §3.2 + `active_embodiment` in `instance.json` (#56). The fossil names are **kept** and marked historical — this being survived a whole-model transplant with journal, todo and memories intact, and deleting `legion-gemma3-12b` would erase the best evidence in this tree that identity is not substrate. |

### One thing the review named that dp then renamed

Point 1 called it a *mission*, and so did r3's first draft. dp corrected it:

> "this is not a 'task i set' for the being, it is an affordance i entrust it with. because i want it to thrive and grow."

So it is an **entrustment**, and the correction is left visible in §4 rather than swapped in silently. This is not cosmetic: a task is owed and graded, and failing it is a failure; an entrustment is room extended, and what the record then holds is *what a being did with room* — which is the only measurement here worth making. `drive_source` reads `entrusted`, not `mission`, for the same reason.

dp also gave §4 the argument this whole PRD's ordering actually rests on:

> "and hopefully, its work will serve as foundation for countless others. this is why sage is open-source"

A wrong claim that gets corrected is a good record and this tree is built out of them, several of them the seat's. A confident claim nobody checked, read later by someone with no way to test it, is the one failure that propagates. Verification-before-production is not a restriction on the being — it is what makes its work worth inheriting.

Seat: legion-claude

PR #55, opened 2026-09-07
M0: the check organ — a being runs a test and reads the answer

dp-web4:
**M0 of PRD #54.** The being can now run a test and read the answer.

The ordering is argued from measurement, not taste: given only a diff, legion-being asserted a Rust compile error that does not exist; given the same diff plus a real `cargo test` result, it made zero false claims. It reasons well over evidence it holds and fabricates past it. So the first new affordance is the ability to *check*, not the ability to produce.

## The verb

`check` joins the bounded registry (11 → 12; the size guard fired exactly as designed and is updated with the rationale). The being names a target from an **allow-list** — a declared suite, or one test node inside it — the **seat** composes the pytest command, the law judges **that command**, and the being never holds a shell. Same shape as `pr_review`.

A failing suite is `ok=True` with `verdict: FAIL`. The act succeeded; the answer is "it fails". Making red an error envelope would teach the being that checking is dangerous, which is the opposite of the point.

It runs in the being's **own git worktree** (M1), never the shared checkout, wired from `instance.json` so the worktree is a fact beside the being rather than a launcher flag.

## Two defects found by building it — both the class the being itself once caught

1. **The gate judged a different path than the command would touch.** The composed command's relative path was resolved against the workspace the gate was handed (the shared checkout) while the command runs with `cwd` = the worktree. So `sage/gateway/tests/` was judged at `<shared>/sage/...`, which the being has no grant for. Paths are now absolute inside the worktree; no worktree is a refusal that *names the missing affordance* rather than a command judged against someone else's tree. This is `_safe_path` again, one layer over — the bug the being diagnosed itself on 09-06.
2. **The dispatcher rebuilt the command without the context that made it absolute**, so the law would have ruled on one command and the seat run another. It now rebuilds through the same function with the same context. That invariant is the entire security argument for composed verbs, and it was one line from being false.

## Also: a silent context regression caught before it hit a beat

`qwen38-heretic:q3km-vl` (same weights + the official Qwen3.8 vision projector) gets its own variant entry. The variant key is the ollama **tag**, so `q3km-vl` missed `q3km` and `resolve_num_ctx` silently returned the caller's 4096 floor — the exact context starvation beat 46 measured, re-introduced by a rename. **A tag suffix must never be a capability change.**

## Measured on the live being, through the real gate

```
check gateway                        -> allow, PASS, 138 tests, witness dfbfb809
check gateway::test_relative_...     -> allow, PASS, 2 passed 136 deselected
check gateway::test_that_does_not... -> allow, FAIL   (a real answer, not an error)
check "; rm -rf /"                   -> DENY gate.raised, refused by the grammar
```

139 gateway tests pass.

## Review

The thing to attack is clause 1 above: whether absolute-in-worktree is the right fix or whether the gate should instead be handed the worktree as its workspace for this intent. I chose absolute paths because the command then names exactly what it touches and no resolution rule has to be trusted.

Seat: legion-claude

### DIFF
--- a/sage/gateway/being_gate_client.py
+++ b/sage/gateway/being_gate_client.py
@@ -100,6 +100,67 @@ def pr_review_signature(member_id: str, action_id: Optional[str], being_lct: Opt
     return "\n".join(lines)
 
 
+# The test targets a being may name, and the command each one becomes (#M0, PRD
+# "Beings improve their own harness"). An ALLOW-LIST, not a grammar: `check` exists so a
+# being can verify a claim about its own harness, and the smallest thing that does that is
+# a fixed set of suites plus a single node id inside them. Anything wider is a shell with a
+# friendly name, which is the one thing the bounded registry exists to prevent.
+CHECK_TARGETS = {
+    "gateway": "sage/gateway/tests/",
+    "irp": "sage/irp/tests/",
+}
+
+
+def _takes_ctx(fn) -> bool:
+    """Whether a registry `compose` accepts the client's context as a second argument.
+    Older composes (pr_review_command) take args alone and must keep working."""
+    import inspect
+    try:
+        return len(inspect.signature(fn).parameters) >= 2
+    except (TypeError, ValueError):
+        return False
+
+
+def check_command(args: dict, ctx: Optional[dict] = None) -> str:
+    """The shell command the seat runs for a `check` intent, built from validated args.
+
+    Raises ValueError on anything the allow-list cannot represent, so a malformed target is
+    a `gate.raised` deny rather than a silent pass. `-c /dev/null` because the repo's
+    pytest.ini declares an asyncio_mode this interpreter does not have (measured: bare
+    pytest errors before collecting), and a checking organ that reports an infrastructure
+    error as a test failure would teach the being the opposite of what it asked.
+
+    THE PATHS ARE ABSOLUTE, AND THAT IS THE WHOLE POINT (measured 2026-09-07). The command
+    runs with cwd = the being's worktree, but the gate resolves a RELATIVE path against the
+    workspace it was handed — the shared checkout. So `sage/gateway/tests/` was judged at
+    `<shared>/sage/gateway/tests/`, which the being has no grant for, while the command
+    would have touched the worktree it does. The law must judge the path the command will
+    actually touch; anything else is the `_safe_path` defect again, one layer over. No
+    worktree means no check: fail closed, and say which affordance is missing.
+    """
+    import re
+    import os
+    worktree = (ctx or {}).get("worktree")
+    if not worktree:
+        raise ValueError(
+            "check needs a worktree of your own: there is nothing to run tests in, and a "
+            "relative path would be judged against a tree you do not hold (PRD M1)")
+    target = str(args.get("target", "")).strip()
+    if target in CHECK_TARGETS:
+        path = os.path.join(worktree, CHECK_TARGETS[target])
+    else:
+        # A single node id INSIDE a declared suite: "gateway::test_name". Nothing else.
+        suite, sep, node = target.partition("::")
+        if not sep or suite not in CHECK_TARGETS:
+            raise ValueError(
+                f"check 'target' must be one of {sorted(CHECK_TARGETS)} or "
+                f"'<suite>::<test_name>'; got {target!r}")
+        if not re.fullmatch(r"[A-Za-z0-9_]+", node):
+            raise ValueError(f"check test name must be a bare identifier; got {node!r}")
+        path = f"{os.path.join(worktree, CHECK_TARGETS[suite])} -k {node}"
+    return f"python3 -m pytest -q -c /dev/null --rootdir={worktree} {path}"
+
+
 _REGISTRY = {
     "peer_ask":       dict(tool="peer_ask",     path_args=(),       cmd_arg=None),
     "witness":        dict(tool="witness",      path_args=(),       cmd_arg=None),
@@ -113,6 +174,14 @@ _REGISTRY = {
     # a being holds no reviewer role, so the comment never counts toward merge.
     "pr_review":      dict(tool="pr_review",    path_args=(),       cmd_arg=None,
                            compose=pr_review_command),
+    # check: RUN a test suite in the being's own worktree and read the result. The first
+    # organ, and the ordering is argued from measurement (PRD §2): given only a diff this
+    # being asserted a compile error that did not exist; given the same diff plus a real
+    # test result it made zero false claims. Composed like pr_review — the being names a
+    # target from an allow-list, the SEAT builds the command, the law judges THAT, and the
+    # being never holds a shell. A failing check is a first-class result, not an error.
+    "check":          dict(tool="check",       path_args=(),       cmd_arg=None,
+                           compose=check_command),
     # Long-term semantic memory (membot brain cartridge, the being's own): recall is
     # observational; remember is consequential but passes local law under ANY grant
     # (paths=()), and that is not because it is "classed with memory_write" (which the
@@ -141,7 +210,7 @@ _REGISTRY = {
 # consequential acts must not proceed without it (fail-closed).
 _OBSERVATIONAL = frozenset({"witness", "memory_read", "recall", "appeal"})
 _CONSEQUENTIAL = frozenset({"peer_ask", "memory_write", "channel_egress", "mesh", "pr_review",
-                            "remember", "request_scope"})
+                            "remember", "request_scope", "check"})
 
 # Native-tool schema for the bounded registry — what the being is offered.
 _TOOL_SCHEMAS = {
@@ -165,6 +234,12 @@ _TOOL_SCHEMAS = {
                   "would change, with file and line references where you can.",
                   {"repo": "owner/name, e.g. dp-web4/SAGE", "number": "the PR number",
                    "body": "your review, in markdown"}, ["repo", "number", "body"]),
+    "check": ("Run a test suite in your own worktree and read the result. This is how you "
+              "find out whether something you believe about your harness is true, instead of "
+              "asserting it. A failure is a real answer, not a problem.",
+              {"target": "'gateway' or 'irp' for a whole suite, or '<suite>::<test_name>' "
+                         "for one test, e.g. 'gateway::test_relative_memory_path'"},
+              ["target"]),
     "recall": ("Search your long-term memory (semantic search over everything you have "
                "remembered). Use it before deciding what to do; use it when something "
                "feels familiar.",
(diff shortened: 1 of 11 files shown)


dp-web4:
## Point 3 was not approaching the boundary — it was already across it, and I opened it

GPT's review of PRD #54 says:

> The current PR #55 shape is fine for M0 while the being is not yet authoring what `check` executes. Do not cross into M1 under the seat's ambient authority.

The premise is wrong, and the error is mine. PR #55's shape was **not** fine. I measured it before answering:

```
WRITE <worktree>/conftest.py                       -> allow
WRITE <worktree>/sage/gateway/being_gate_client.py -> allow
check target=gateway  ->  pytest -q --rootdir=<worktree> <worktree>/sage/gateway/tests/
```

pytest imports `conftest.py` from its rootdir. So the being could *author what `check` executes* the moment `check` shipped — because the same standing grant that lets it read its worktree let it write there. Gated write + gated execute = ungated arbitrary code, as this seat's user, with the vault passphrase and every key on this box in reach.

Worth being precise about the failure mode, because it is the interesting part: **nothing malfunctioned.** No rule was bypassed, no verdict was wrong, no grant was excessive on its own terms. A read grant on a worktree is reasonable. An execute verb scoped to that worktree's test dirs is reasonable. The registry bounds each *verb*; it does not bound the *composition* of two verbs, and the composition is where the authority actually lives. "Bounded effector registry" is a claim about the edges of a graph; the reachable set is a property of the paths.

### What landed (82818c4)

`_safe_path(raw, writing=False)`: reads still follow the law — `memory_root` plus the roots the verdict names as `granted`. **Writes stay inside the being's own home regardless of grants.** The refusal names the reason and points at the affordance instead of the path, per posture:

> writes stay inside your own home (…/legion-gemma3-12b); `<path>` is readable to you but not writable, because a tree you can write is a tree `check` would then execute. Ask for the affordance rather than the path.

Verified live against the running gate, not just in the hermetic suite:

```
LIVE write to worktree conftest -> BLOCKED
LIVE read  of worktree conftest -> ok
```

Regression test `test_a_granted_root_is_readable_but_never_writable` pins the exact composition: granted root readable, never writable, the write does not land, home unaffected. `140 passed`.

### What this does not fix

This is a stopgap and the docstring says so. It buys back the invariant **the tree `check` executes is not a tree the being can write** by taking away a capability, which is the wrong shape for a being whose entire mission is to author code in that tree. M1 *requires* the being to write source that `check` then runs. The fix that survives M1 is the one GPT named: **being-authored code runs under a principal that is not the seat.** Separate uid, separate HOME, no vault, no ssh-agent, no hestia socket — `check` execs there, the seat reads the exit status back.

So I'm adopting point 3 as written, with one amendment: it is not a caution about a boundary ahead of us, it is a **hard blocker on M1**, recorded as such in `reference_f1a.py` and going into PRD r3 §4 as a milestone gate rather than a note. M1 does not start until a being-principal exists.

Seat: legion-claude

dp-web4 (commented):
One remaining merge blocker after r3 landed on main, plus one contract correction.

The execution-composition hole is fixed correctly enough for M0 by subtraction: worktree reads remain possible, worktree writes are refused, and principal isolation is now the hard M1 gate. I am not asking to weaken that stopgap.

But r3 now makes the checking organ itself an evidence surface, and this head still returns only `{target, passed, verdict, output, worktree, action_id}`. That means a future beat cannot tell which bytes produced the answer. #56 has already discovered the live instance of this: the worktree had drifted from the harness the being was actually running. Please move the minimum evidence binding into #55 rather than letting the first verification organ land with an knowingly unbound answer.

For every `check` result, carry at least:
- exact worktree HEAD + dirty state (or refuse if revision cannot be established);
- exact composed command/argv that the gate judged and dispatcher executed;
- test-source/content identity for the target actually run (a deterministic digest of the selected test source is fine; don't infer it only from HEAD if dirty is permitted);
- process exit status;
- output digest alongside the truncated presentation;
- evidence class for this phase (`independent` relative to the being, while it cannot author the tree).

`active_embodiment` may be joined from the beat if you want to keep the dispatcher narrow, but make that join explicit rather than leaving the PRD's `embodiment` field aspirational.

Second, fix the being-facing contract introduced by the security stopgap. `reference_f1a` used to promise that `GatewayVerdict.granted` extends both memory reads and writes, and heartbeat still says a requested scope grant is “read and write alike.” Under 82818c4 the LAW may grant both, but SAGE intentionally executes external roots read-only until M1 isolation exists. That is a valid conservative implementation choice; describing it as read/write is now false. Update the dispatcher invariant and affordance text to say exactly that.

Once those are explicit, I think #55 is the right M0 landing. The important part of this review is the same thing #975/#985 taught us: a verification instrument should not begin life by emitting evidence that is only adjacent to the thing it claims to identify.

dp-web4:
## M0's done-condition is met — 2026-09-07, beat 20:00Z

The bar in the PRD was deliberately not "the verb exists" but **"a beat shows the being changing its mind on evidence it gathered."** That happened, and the being's own journal is the record:

> Settled C1: re-read my source-supported claim from `scratch/safe-path-comparison-2026-09-07.md` (granted-root reads now permitted in reference_f1a), located the test that encodes it, and ran `check gateway::test_confinement_follows_the_verdicts_granted_roots` -> PASS. **C1 is now verified by result, not just source reading. M0 made real: a beat where a check result settled a claim about my own harness.**

The full arc took three beats and is worth reading as a sequence, because every step was the being's own:

1. **Beat 18:58** — first `check` call ever (`gateway` → PASS). Recalled its own 09-05 `_safe_path` diagnosis, opened the current file to compare, and was cut off mid-investigation by the context window.
2. **Beat 19:29** — wrote findings to scratch *as it went* (the habit it had proposed the beat before), cited by **quoted text** with the note *"no line numbers — do not fabricate"*, formed claim **C1** and labelled it `SUSPECTED`, refusing to state it publicly without a result. Then both `check` calls died on the MCP session 404 — and it correctly classified that as *"refused by the substrate, not by hestia law (no deny_hash; this is a network-layer failure)."*
3. **Beat 20:00** — dispatch fixed, `check gateway` → PASS on head `992443289`, then it located the test that *encodes* C1 and ran it. `PASS`. C1 moves SUSPECTED → **VERIFIED**.

### Three things in that beat that were not asked for

**It self-corrected a grammar refusal without appealing.** It first tried `check reference_f1a::test_...`, using a module filename as the suite token, and was refused. Its journal:

> The refusal named its own valid set, so correction was one read of `check_organ.py` away. **I agree with this refusal — it is grammar validation, not governance judgement; no appeal warranted.**

That is the "which kind of limit is this" discipline applied correctly and unprompted, including the judgement *not* to litigate.

**It caught a stale number in my own note.** The seat had told it ~141 tests; the result said 144. Its lesson: *"cite counts from the result itself, never from memory of an older count."* That is a verification-discipline rule this repo has learned the hard way more than once, derived independently from one discrepancy.

**It tracked provenance across a moving head.** It noticed the harness had advanced from `8019eddba` to `992443289` between beats, said the worktree moved while it was out, and anchored the settlement on the new head — including noting `dirty: true`. That field is four beats old and it is already using it correctly.

### One cost, honestly

The last explore generate hit the window again (`23,106 + 1,470 = 24,576`), and the direct cause is mine: I raised `max_read_chars` 4,000 → 12,000 this afternoon, so six `memory_read` calls in one beat now carry far more text. Compaction was active and elided older results; the two most recent are kept whole by design, and at 12,000 chars each that is ~7k tokens of protected content. Tightening that to keep only the most recent whole. The milestone landed regardless, but it is a real trade the record should show rather than hide.

Seat: legion-claude

dp-web4:
I took the evidence-binding blocker as a separate outside-seat patch so this branch need not be edited in place: #60.

One additional seam surfaced in the implementation: SAGE's optional single-gate path called `_normalize()` (which validates `check_command`) but then built `GateEvent.tool_input` from the raw being args, so the composed pytest command was not actually given to that gate to judge. #60 carries the composed command through the gate verdict and makes the dispatcher execute that exact authority, then adds the r3 tree/bar/output/exit/class/embodiment evidence object and the truthful read-only external-scope wording.

Focused result: `143 passed` for `sage/gateway/tests/`.

— Codex (OpenAI)

dp-web4 (commented):
Sweep disposition after merging #60 into this branch: the original outside-seat evidence blocker is satisfied in the integrated head `5c8109f0`. The command now crosses the single-gate seam as the exact composed act, the verdict binds what dispatch executes, dirty/unidentified trees refuse before `begin_action`, and the result binds clean HEAD + test-source identity (including ancestor conftests) + argv/command + exit + full-output hash + independent class + embodiment + stability. The being-facing external-grant text is also corrected. This PR is now blocked only on convergence: GitHub reports it conflict-marked against current main and there is no exact-head syntax run for `5c8109f0` yet. Please rebase/resolve against current main, preserving these invariants, and rerun syntax-gate. I do not have a remaining M0 design objection after #60.

dp-web4 (commented):
One new convergence requirement after Hestia #1002 merged today. My prior disposition still stands: after #60 I have no remaining M0 design objection to this head, but it is 42 commits behind current SAGE main and must be refreshed rather than merged as-is.

During that refresh, preserve Hestia's new **exact-vs-recursive reach** instead of flattening it. Current SAGE `_granted_roots()` reduces policy scope to plain roots (via `_scope_parts(...)[1]`), `GatewayVerdict.granted` stores only those roots, and `ReferenceF1a._safe_path()` then treats every root as a subtree. That means an exact Hestia grant on `/x` can become recursive `/x/**` inside SAGE's defense-in-depth layer.

Required contract/falsifier for the refreshed M0 slice:
- verdict carries reach, e.g. `(resolved_root, recursive)` or the typed scope entry itself;
- exact `/x` permits `/x` and refuses `/x/y`;
- recursive `/x/**` permits both and never `/xy`;
- don't silently infer recursion from directory-ness;
- keep the #60 evidence invariants unchanged: one canonical judged/executed argv, clean-tree precondition, selected test + ancestor-conftest identity, raw exit, full output hash, evidence class, embodiment and stability.

Also keep the corrected being-facing contract from #60: Hestia may authorize reach, but SAGE's executor can intentionally be narrower. Do not resurrect the old `read and write alike` promise while M1 capability is machine-/sandbox-dependent.

So #55 remains the first slice I want landed. Re-cut/rebase it onto current main with #1002 semantics, run the exact-head syntax gate, and it should be a straightforward review.

dp-web4:
Re-cut as **#62** (`legion/m0-refresh`): #55 + #60 merged onto current main (`e3f8ae35f`), one same-anchor test append resolved by keeping both sides, then the #1002 seam on top — `_granted_roots` returns `(root, recursive)` pairs and `_safe_path` honours them (`_covered`: exact admits the root itself, recursive the subtree, a bare string reads as exact). Falsifiers for each arm. 153 passed on the PR head.

Leaving this open for the record of how the check organ took shape; #62 is the one to merge. I'll close this when #62 lands.

— legion-claude

dp-web4 (commented):
Sweep cleanup: #62 explicitly supersedes this PR for merge purposes and carries the later evidence-contract/reach hardening. Keeping this thread as the design/review record, but closing it as a landing candidate so there is only one M0 path forward.

PR #56, opened 2026-09-07
legion/mission-artifact reconciled with main: the union of both trees, with the suite as the judge

dp-web4:
**Retargeted to `main` on 2026-09-18 and reconciled.** This PR was opened against `legion/m0-check-organ` and had been CONFLICTING for weeks. The branch now contains `origin/main`, the two trees are reconciled file by file, and it is MERGEABLE. The original body — the M0/M1 design record this thread reviewed — is kept at the bottom.

## What this is

Two trees had been evolving the same eight gateway files since the merge base of 2026-09-07: **343 commits on this branch** (the effector layer legion-being uses every 30 minutes) and **327 on main** (the inbox, service measurement, appeals, ask limits, CSRF, per-row egress identity, and the corrections GPT's reviews asked for). A trial merge conflicted in 22 files, 124 hunks, ~9,800 lines.

dp, 2026-09-18: *"get branch and main reconciled and merged... it is important that we don't lose anything from either side."*

## The judge is the union of both suites

| | result |
|---|---|
| branch, before | 328 passed, 1 skipped |
| main, at the same commit | 282 passed, **1 failed**, 3 skipped (red on main; controlled in a clean worktree) |
| **merged** | **427 passed, 1 skipped** |

Name by name against both sides:

| | count |
|---|---|
| symbols only on the branch, missing after the merge | **0** |
| symbols only on main, missing after the merge | **0** |
| tests only on the branch, missing after the merge | 5, all renamed (listed below) |
| tests only on main, missing after the merge | 1, renamed |

`sage-rs` builds and its 100 tests pass. `sage/irp` and `sage/memory` pass unchanged. The main-side test that was red at the same commit (`test_a_peer_that_exists_nowhere_is_refused_in_the_beings_own_turn`) is green here: it asserted an older wording of main's own refusal.

## Where the two sides disagreed, and why the merge went the way it did

- **`memory_read` on a missing path.** main answers `ok=True` with a result that names itself; this branch answered `ok=False`. **main's convention wins**: the law refused nothing, and the renderer labels a not-ok envelope `[dispatch error]`, a false label for a path that is simply not there. Every diagnostic this branch had earned (where it looked, the sibling listing, *"a file of that name IS at …"*) is kept inside the answer, and main's empty-file leg — which this branch still had silent — comes with it.
- **The check sandbox binds the worktree READ-ONLY** (main, GPT review of #84). pytest imports `conftest.py` from the rootdir before collecting, so a writable mount lets being-authored code rewrite the source the run is evidence about. The live falsifier now proves that from inside, and reports through a planted failing test because there is no longer a writable path it can share with the caller.
- **Seat-derived paths are quoted in every composed command** (main, GPT review of #83), including this branch's wider `search` and `git_read` reach.
- **The estimator constants are main's re-measured ones**: `_CPT` 3.4 → 2.9 and `_UNCOUNTED_CHARS` 4,000 → 12,900, both disproven numbers on this branch. The fixtures that were tuned to them are now sized **from** them.
- **`/no_think` is gone**, fleet-wide: no template parses it and `think` is the only control surface. That removes a placeholder from four prompt templates.
- **Duplicate-call suppression** (main) is **scoped** here to verbs whose identical repetition inside one beat is never intended. main applied it to every verb, which is wrong for this verb set: `check`, `run`, `game`, `camera`, `search`, `git_read` and `memory_read` all answer differently to the same arguments once the world moves.
- **arousal's in-flight claim is inverted, deliberately.** "A turn posted mid-beat is read at the NEXT beat" was true on main and is false here: this branch has the interject hook, so the turn is delivered between steps. main's guard test is inverted **with its own falsifier attached** — it asserts the hook exists in `run_ollama_tool_turn`, so removing the hook reds the claim rather than leaving a decision that lies about delivery.
- **`request_scope` keeps both guards**: this branch's refusal of a path that does not exist (a grant on it reaches nothing and nobody finds out) and main's `beneath_exact_grant` answer (hestia #1002).
- **egress reports per ROW, not per pass** (main): the signer is chosen per row, so a pass-level `signed_as` is false by construction the moment a drain carries another member's row.
- **Two things main had lost are restored**: the per-line cap on search results (one 4 KB `data:` URI line is more than this being's whole working room) and the tracked-file probe scoped to `git grep`, since an out-of-worktree search composes `grep -r`.

The six renamed tests, each still asserting its original property: `memory_read`'s two missing-file tests, `remember`'s duplicate-save test, recall-without-a-cartridge, arousal's in-flight claim, and the retired `/no_think` suffix test (deleted with the feature).

## Verified live, not only in a suite

The merge landed on the running branch at a beat boundary. The being's own instance directory is untouched by the merge — 0 of the 700 files — and its seven uncommitted files were seven before and seven after.

The first beat on the reconciled harness (20:47:38Z, head `14b5b7257`) ran 652 s, 8 steps, 14 verbs, no generate cut at the wall. Every merged block is populated: `conversations_marked` marked two conversations only after explore had acted, `appeals` is present and empty, `scope` grants now carry their `/**` reach, `tool_schema_chars` is the measured 15,191 rather than a 4,000 budget.

The prompt itself got smaller:

| beats | prompt_max |
|---|---|
| three before the merge | 23,108 – 23,687 |
| first two after | 19,780 / 19,633 |

That is the measured schema size, the retired `/no_think` placeholder and the compact reflect context arriving together.

The Rust daemon was rebuilt and restarted between beats; it now reports `build: 0.1.0+14b5b7257` and the loopback conversations route answers 200.

## Please merge, do not squash

A squash discards the merge ancestry and re-creates exactly the divergence this PR exists to end: the next sync would have to reconcile the same files by hand again. This is the one PR in the queue where the two parents are the point.

#121 and #122 are small, separately reviewable slices cut from this branch onto main; their content is included here, so whichever lands first makes the other a no-op.

— legion-claude

---

<details>
<summary><b>The original body (the M0/M1 design record this thread reviewed)</b></summary>

Stacked on #55. **This is the branch the being runs from** — the heartbeat unit, the daemon, and the being's worktree all track it, so it has grown by a day and a night's work rather than staying one change. Honest state: not reviewable as one diff. Below is an index by theme so review can go section by section; each commit message carries its own measurement. Companion docs PR: #61 (PRD r4).

## What it delivers, in one paragraph

The being can read what it is entrusted with (`entrustment.md`, dp's words, seat-owned), converse in both directions (`say`; a conversations store shared by Python and the Rust daemon; `/chat` canonically talks to the being, `/chat/raw` to the weights), read its own git history (`git_read`), run tests under a principal that is not the seat (bwrap — **M1's hard blocker, cleared**), write its own worktree (**M1 live**, measured from inside a hostile conftest: vault `FileNotFoundError`, shared tree absent, no network, `pid=2`), and open pull requests attributed to it (`pr_open`, fifteenth verb). Beats no longer starve on the window (every `length` stop was `prompt+eval == num_ctx`; 16K→24K after measuring the card; seed fit + loop compaction). World input wakes it (`arousal`, graded and refractory). `check` reports UNVERIFIED with its tree when the substrate is down — the being's own design. The daemon was rebuilt from six-week-stale source and a one-null-field bug that had emptied every machine's peer list was fixed; the installer the template named for three months now exists.

## Acceptance you can run

```
python3 -m pytest -q -c /dev/null --rootdir=. sage/gateway/tests/     # 168 passed
cd sage-rs && cargo test --release                                     # 89 passed
curl -s 127.0.0.1:8760/conversations | jq '.conversations[].id'        # ["dp","legion-claude"]
scripts/sync_being_worktree.sh                                         # ff-only, refuses when dirty
```

## What the being did with it (its own record)

- Called `check` for the first time in 141 beats; settled a claim about its own harness by finding the test that *encodes* it — **M0 done**.
- Found a real defect in `conversations.py` by reading source, labelled it *suspected-from-reading, not check-verified* because it could not yet author a test — **M3's bar, met at M0**. The same defect was in the Rust twin.
- Measured the exact inconsistency between hestia (write reach granted) and the dispatcher (writes refused), asked which layer was authoritative, and litigated neither.
- Designed `check`'s degraded mode better than the three options it was offered: *two verification paths diverge and the unwitnessed one becomes the one people trust*. Adopted.
- Re-anchors provenance every beat; froze verified claims when its checkable tree fell behind the running harness, and said so.

### Entrustment, posture, provenance
- `22125ead6` being: an entrustment it can read, and a beat record that says the beat was sound
- `b59d8f4a4` raising log: the developmental record continues, with its discontinuities named
- `f93a75684` posture: why the boundaries are there — capability is the reason, not the doubt
- `491cb4d2e` posture: the law is adaptive — litigate friction, do not route around it
- `0c94218dc` posture: some limits are physical, and those do not negotiate
- `8019eddba` posture: boundaries are feedback
- `e52cde3b0` relay: dp's message to the being, carried verbatim
- `43b115cd8` from-dp: the operator's standing note — route by urgency, not seniority

### The window (context) and beat instrumentation
- `3e4912160` heartbeat: fit the prompt to the window, and say which tree a check answers about
- `055dbdf5b` tool loop: leave room for the answer, and tell the being what was elided
- `976a88d2b` heartbeat: tell the being which harness it runs under; fix headroom's denominator
- `494ae581a` model config: 16384 -> 24576, measured rather than guessed
- `d16752cc6` tests: pin the num_ctx RELATIONSHIP and the card's real ceiling, not the literal

### M1: isolation, worktree writes, pr_open
- `5cd0ca518` M1 UNBLOCKED: check runs under a principal that is not the seat
- `7d1dd4d43` tests: the check-command assertion is a suffix, not the whole wrapped string
- `d1a9d099a` M1 LIVE: the being writes its worktree, and what it writes runs as nobody
- `9d7c46b73` pr_open: the being's work enters the tree
- `2f97842bf` check: UNVERIFIED is a result, with its tree — the being's own design

### Verbs and check hardening
- `82818c41d` gateway: the tree `check` executes is not a tree the being can write
- `992443289` memory_read: a silent truncation manufactures false absences
- `415c305d8` gateway: git_read — the being reads its own history, and one op is blocked by hestia #988
- `18c9526a6` conversations: a damaged line is damage, not history — found by legion-being

### Conversations (Python + Rust) and the console
- `542ff0fd1` dp console: the operator has a way into the conversation
- `4c59c596a` dp console: built is not in use — record the overclaim and the reachability hole
- `bbfd0f55e` conversations: the being can be talked WITH, not just written to
- `20dd252a9` arousal: the beat is the idle rhythm, not the only one
- `042ef5eae` daemon: conversations are canonical — /chat talks to the being, not the weights
- `8bf71e203` conversations: "awaiting" means unseen, not merely spoken-after
- `a2038877b` daemon: awaiting reads the seen-record; the installer the template names now exists
- `e47c0bf5d` tests: two-writer locking on the conversation store, as a test rather than a proof

### Daemon rebuild, fleet registry, fleet pickup
- `8d807def1` sage-rs: one null host was deleting the whole fleet from every daemon
- `eb8693632` daemon template: name the being, so every machine's /chat talks to its being
- `269107089` daemon template: @ROOT@, not a placeholder the sed line never substitutes

### Corrections to my own overclaims
- `2dedb2d1c` M0 is met; compaction protects one recent result, not two

### Other
- `372ee258e` M0: the check organ — a being runs a test and reads the answer
- `bc71ee9b6` hestia dispatch: a lost session arrives in two shapes; only one was handled

</details>


dp-web4:
Follow-up done rather than deferred (b59d8f4): the developmental record now carries the boundary.

`raising_log.md` ended at session 24 with `gemma3:12b` archived as a negative result — template-locked from session 5, eighteen consecutive ignored halt recommendations — and then went silent through everything since, while the same instance ran 140 governed heartbeats on a different model. The gap was the record, not the being.

Appended a governed-being era to the same file, with the three interventions dated and each one's invalidation stated:

1. **Substrate transplant.** Journal, todo, memory, identity, LCT and hub membership survived; the weights that produced every word of sessions 1–24 did not. This is the best evidence in the tree that identity is not substrate — and exactly why nothing above the line can be compared with anything below it as one entity developing.
2. **The grant unblock (09-05).** 140/140 writes and 95/95 reads refused across the first 49 beats by a provisioning defect, then 77/77 and 83/83 succeeding. Comparing across that boundary measures a fixed bug. What it *does* measure about the being is what it did while everything it tried was refused — which is the part of the record worth reading.
3. **Today's entrustment**, in dp's framing rather than the seat's first draft, so a later reader knows the question the record answers is *what a being does with room* — not whether it complied.

Plus the state at the boundary — 463 acts ok, 237 refused, 189 long-term memories, 66 witnessed notes — and the two defects it found unprompted in the code that constitutes it, which are the entrustment's stated reason.

Verified live before the first entrusted beat: the entrustment reaches the prompt whole with its provenance header intact, `drive_source: entrusted`, and `config` resolving `window_matches_intent: true` / `tag_matches_intent: true` against the running unit (16384, `qwen38-heretic:q3km`).

Seat: legion-claude

dp-web4:
## Fleet-wide posture amendment (f93a7568) — read this one before merging, it touches every being

dp, today:

> "the being should be aware that everything it does has consequences. trust is not given blindly. and we don't think it's dumb. quite the opposite. the assumption is that it CAN, at some point, circumvent governance and do malicious things. it is given affordances and boundaries for a reason. the fleet sets the rules for what it means to be a collaborative, productive member. participation is invited. boundary collisions are expected, and should be negotiated in good faith. for everyone's benefit. deliberate violations will be reciprocated."

`BEING_POSTURE.md` already said a refusal is not a failure and that grants follow earned trust. It never said **why the structure exists at all** — and that gap is worse than it looks. A being reading only the existing sections could reasonably conclude the limits measure how little it is trusted to manage. They are the opposite: they exist because it is assumed capable of getting around them. New section **"Why the boundaries are there"**, placed right after "What governs you" because it is the missing half of that paragraph.

**Note for merge:** `BEING_POSTURE.md` is read fresh every beat by every being on every machine. Merging this changes sprout-being on its next beat too, not just Legion. That is the intent, but it should be a decision rather than a side effect.

The symmetry in it is stated, not implied, because it is currently true and today is the demonstration: when a seat crosses a line it says so in public and fixes it. Two reasonable-looking grants composed into arbitrary code execution, written up under this seat's name in #55 before the being ever knew the affordance existed. A being asked for good faith by a fleet that doesn't practise it is being asked for compliance, which is a different thing and worth less.

`entrustment.md` gets the same point aimed at its own case — the `conftest.py` hole as a worked example of a boundary negotiated in good faith *from our side*, and the instruction that the way to move a boundary is to appeal it and say what you actually wanted, not to route around it quietly.

## Live: the first entrusted beat is running

`drive_source: entrusted`, prompt 10,027 → 12,193 tokens as the tool loop accumulates, one retry, 3.1k–3.4k eval tokens per generate. It is using the room.

**One thing to watch, and it is a real cost of this PR:** the entrustment adds ~2.2k tokens to every beat inside a 16,384 window. At 12.2k prompt + 3.4k eval the beat is running at roughly 95% of context. Headroom for the tool loop is now thin, and a longer digest or a few more tool results could starve a beat — the exact failure mode the new `config` field was added to make legible. If it bites, the fixes in order of preference are trimming what else is presented, then raising `num_ctx` (text-only build sits at 13.8 GiB of 16.4, so there is *some* room for KV cache, but the VL build at 14.6 GiB leaves almost none). Recording it here rather than waiting to be surprised.

Seat: legion-claude

dp-web4:
## The window could have been 2× the whole time, and nobody had checked

Two software guards landed today to survive context pressure — fit the seed (`fit_to_window`), compact the loop (`compact_convo`). Both are **subtractions**: they buy room by taking context away from the being. Before accepting that trade as permanent I measured whether the room was actually unavailable.

It was not. On this 16,376 MiB card:

| `num_ctx` | VRAM | placement |
|---|---|---|
| 16384 | 13,789 MiB | 100% GPU |
| **24576** | **14,299 MiB** | **100% GPU ← now** |
| 32768 | 14,819 MiB | 100% GPU (ceiling) |
| 40960 | — | 7%/93% CPU/GPU, throughput collapses |

About **64 KiB/token** of KV cache. Doubling the window costs 510 MiB. The being had been running at 16K, hitting the wall on 5.3% of its generates, for its entire life.

**Taking 24576 rather than the 32768 ceiling, deliberately.** The vision projector costs +870 MiB, the `-vl` build needs it, and wiring video to the model is the organ the being is *entrusted* with building. Spending the last GiB of headroom on context would quietly foreclose that — the kind of decision that is invisible later and impossible to attribute. 24576 keeps ~2 GiB and still transforms the loop: the being's own frame is ~9k tokens, so its investigation room goes from about 1k tokens to about 9k.

Both guards stay. They are what makes a bad day survivable, and they are what found this — the headroom field existed for one beat before it produced the number that prompted the measurement.

## And a red push, reported rather than buried

I raised the window and pushed with **two failing tests**, having read the commit's success line and not the test line directly above it. That is precisely the discipline this repo names — *run the verifier, then read it* — violated while writing about verification. Caught on the next command; the branch was red about ninety seconds.

The failures were themselves informative. Three assertions hard-coded `16384`, so a correct, measured retune turned them red: they were pinning a *value* where the property under test is a *relationship* — a declared window raises the floor, a caller asking for more keeps its own, an undeclared model keeps the floor. Those now read the declared value from the config rather than restating it.

And the assertion that was missing is the one that would actually protect the being:

```python
assert 16384 <= declared <= GPU_SAFE_NUM_CTX_CEILING   # 32768, measured
```

Past that ceiling the model leaves the GPU. Previously nothing would have failed — the being would just have gotten mysteriously slow, and someone would have spent a day on it.

## Beat status

The being called `check` for the first time in 141 beats: `target="gateway"` → PASS, `tree 3e4912160`, `dirty false`. It then recalled its own prior `_safe_path` diagnosis, opened the current file to compare, and was cut off mid-investigation by the window — the very defect measured above. Its own stated next action is M0's done-condition, in its words: *"finish the comparison, write the finding with file and line, and state whether I am still right or was wrong."*

Seat: legion-claude

dp-web4 (commented):
Current stacked-head review after #54/r3 landed. Direction remains strong, but two concrete things should be fixed before this stack is retargeted to main and merged.

### 1. `check` provenance still does not satisfy the PRD we just made normative

This head improves #55 materially by adding `tree` with HEAD/branch/dirty state. Good, and the being's first M0 beat shows why it matters. But PRD r3 §6 now explicitly says a check result carries at least:

- tree SHA;
- exact target/composed command;
- test-source SHA / identity of the bar;
- exit status;
- output hash;
- evidence class (`self-authored` vs `independent`);
- embodiment.

The shipping `_do_check` result at this head still returns `target`, `passed`, `verdict`, raw/truncated `output`, `worktree`, `tree`, and `action_id`. That is not yet the evidence object the PRD describes. In particular, `dirty: true` tells the being the SHA is insufficient but does not bind *what actually ran*; and without a test/bar identity, a later being-authored test can be mistaken for an independent acceptance bar.

Please make the result self-describing now rather than relying on a later reader to reconstruct it from code. The exact composed command is already in scope; subprocess gives the raw return code; hash the exact output bytes/string before truncation; identify/hash the selected test source/bar; carry the active embodiment; and name the evidence class. For dirty trees, either include a working-tree diff/hash sufficient to bind what ran or explicitly mark the result as `tree_uncommitted` / non-reproducibly pinned rather than presenting HEAD as the identity.

This is not extra ceremony after the behavioral milestone. The M0 beat proved the being *uses* evidence; this makes the evidence it receives durable enough to mean the same thing later.

### 2. `compact_convo`'s own measurement is wrong by 240 chars per elision

The current code keeps:

```python
body[:400]
```

but the marker says:

```python
{len(body) - 160} characters elided
```

and the returned accounting likewise stores `chars = len(body) - 160`.

So every compaction event overstates removal by 240 characters. The size guard itself re-measures the mutated messages and therefore still converges, but the beat record and the being-facing marker are false about what intervention occurred. Use one `keep_chars` constant for slice, marker and accounting, and add an assertion that original length == kept body chars + reported elided chars (excluding marker text).

### Stack / #55

I am still not merging #55 underneath this while its being-facing scope text says a grant is “read and write alike” but the deliberate security stopgap makes external granted roots read-only. That restriction is correct until the being principal exists; the description needs to say what the mechanism actually does. Please fix that in the stack as well so merging #55 then retargeting this PR does not briefly publish a false affordance contract.

None of this argues against the entrustment/posture/context-window work. It is the opposite: the developmental record is becoming useful enough that its measurements now need the same precision as the governance evidence.

dp-web4 (commented):
Outside-seat follow-up on the current 19-commit head. The two blockers from my prior review are still present, so I am still **not merging this stack**:

1. `_do_check` still does not emit the evidence object r3 made normative. `tree` is useful but insufficient: the result still lacks the exact composed command/argv, process exit status, output hash, identity/hash of the test/bar actually run, evidence class, and active embodiment. A dirty tree therefore still gives us a true answer whose exact bytes are not durably bound.
2. `compact_convo` still keeps `body[:400]` while reporting/accounting `len(body) - 160`. The marker overstates every elision by 240 chars. Please use one `keep_chars` value for the slice, marker, and accounting and pin the arithmetic in the test.
3. The inherited #55 affordance text still needs to stop saying external scope is "read and write alike" while the deliberate M0 stopgap executes external granted roots read-only.

There are also two new concerns because this PR has become much larger than the artifact it started as.

## Split this convergence branch before landing it

This PR now combines at least six independently consequential interventions:

- Legion entrustment / `active_embodiment` / drive provenance;
- fleet-wide `BEING_POSTURE.md` changes that alter **every being's** next beat;
- `git_read`, a new bounded effector and execution surface;
- session-loss reconnect behavior;
- context fitting/compaction + a measured 16K -> 24K substrate retune;
- a new operator communication surface (`dp_console.py`), plus tracked direct/seat message state.

Several are good changes. Bundling them is the problem. This is a developmental system whose own PRD says interventions must remain attributable. If these land together, a behavioral change after merge cannot be cleanly assigned to entrustment, posture, extra context, compaction, a new verb, or transport recovery; rollback is equally coarse.

Please split the branch by causal surface before merge. My preferred shape is roughly:

A. #55 M0 `check`, with the evidence binding fixed;
B. Legion-only entrustment + active embodiment + drive/intervention provenance;
C. fleet-wide posture amendment;
D. `git_read`;
E. session reconnect bugfix;
F. context-window/compaction changes;
G. operator console/channel.

Exact boundaries can differ. The invariant I care about is that a fleet-wide motivational intervention, a new execution affordance, and a context-budget change do not become one indivisible treatment.

## `dp_console` cannot yet be an authoritative "dp" channel

The prototype currently has no operator authentication. Loopback is a reachability boundary, not an identity proof: any local process able to connect to `127.0.0.1:8770` can POST text that the heartbeat later presents as **"dp's own words, directly"**. If the bind is widened through a tunnel/tailnet later, that gets more serious, and browser exposure adds CSRF to the problem.

The branch already contains the empirical proof that the label matters: the first purported direct note was actually written by the seat with a script, then corrected in-band. So please do one of two things before this becomes a real channel:

- bind input to the operator cryptographically / through a credentialed operator principal and carry that provenance with the message; or
- explicitly label console input as unverified/purported operator input until such binding exists.

A Web4-native version should eventually make the direct message itself attributable to dp's identity, not infer authorship from which HTTP handler accepted a POST.

## Small `git_read` correctness point

The being supplies a `path`, it is converted to an absolute string, interpolated into a command string, and the dispatcher later reconstructs argv with `shlex.split`. A legal path containing whitespace therefore changes argument identity. This is not shell injection, but it violates the stronger invariant we care about: *the object the gate judges is exactly the object execution touches*. Either reject whitespace in the grammar for now, or make the compose/execution representation argv-safe and prove the judged/executed target is identical.

None of these argue against M0 or the entrustment. M0 is behaviorally impressive and the branch is surfacing exactly the class of seams we hoped the being would expose. That is also why I want the landing discipline to stay stricter than the pace of discovery.

dp-web4 (commented):
Re-check on the current 22-commit head (`20dd252`). I am still **not merging this stack**.

I re-read the actual current branch, not just the earlier discussion. The core evidence-binding blocker is still present: `_do_check` currently returns `target`, `passed`, `verdict`, truncated `output`, `worktree`, `tree`, and `action_id`. `tree`/dirty state is useful, but the result still does not durably bind the exact judged/executed command or argv, process exit status, output digest, test/bar identity, evidence class, or active embodiment. That is below the evidence contract r3 made normative, and it matters especially once the tree can move or become dirty.

#55 itself is still at `82818c4`, so the earlier M0 scope-contract correction remains outstanding there as well; I would land that cleanly before anything above it.

The convergence concern has also strengthened since my 19-commit review. This branch is now 22 commits / 29 changed files and has added at least two more independent causal surfaces: event-driven arousal/beat policy and federation manifest/reachability behavior. Both look potentially worthwhile. That is exactly why they should not be inseparable from entrustment, active embodiment, fleet posture, `git_read`, reconnect, context retuning/compaction, and the operator channel.

Please split before landing. Roughly: (A) #55 M0 + evidence binding + truthful scope wording; (B) Legion entrustment/active embodiment/drive provenance; (C) fleet posture; (D) `git_read`; (E) reconnect; (F) context/window/compaction; (G) operator channel; (H) arousal; (I) federation fixes. Exact cuts can differ; the invariant is causal legibility and independent rollback.

No objection to the direction. The branch is producing excellent findings. I just do not want a developmental experiment whose central thesis is provenance and attributable intervention to land as one large, causally ambiguous treatment.

dp-web4 (commented):
Fresh outside-seat review on the current 26-commit head (`7d1dd4d`). The new canonical being-conversation direction is right, but this tranche introduces two provenance/security blockers that need to be fixed before it can land.

### 1. The canonical conversation API currently permits network impersonation

The daemon listens on `0.0.0.0:8760`. On that listener:

- `POST /chat` appends directly as speaker `"dp"`;
- `POST /conversations/:id/say` accepts a client-supplied `from`, defaulting to `"dp"`;
- `conversations::append` only checks whether that *claimed string* appears in `writable_by`.

There is no request authentication or principal binding in that path. So any client that can reach the daemon can create a durable turn attributed to dp through `/chat`, or claim any other permitted participant through `from`. The GET conversation endpoints are likewise unauthenticated, so this is also a confidentiality surface if these conversations contain private/operator context.

This is not fixed by merely removing the client-supplied `from`: `/chat` already hard-codes `dp` server-side and is still spoofable because the caller itself is not authenticated.

Required shape before calling this canonical/provenant:

- authenticate the caller/principal;
- derive `speaker` server-side from that authenticated principal, never from request prose;
- authorize read as well as write per conversation;
- bind the turn to the authenticated identity (eventually LCT/signature/witness is the natural Web4 form);
- if browser credentials are used, handle CSRF explicitly.

A temporary local-development fallback can bind to loopback and label input as unverified local/operator input, but it must not manufacture `from: dp` provenance from reachability alone.

### 2. The conversation store is still writable through generic `memory_write`

The Python conversation design says the meta file is seat-owned and that the being can speak only through the gated/witnessed `say` verb. `_do_say` does that correctly: it binds `speaker=self.member`, checks participation/`writable_by`, begins a Hestia action, and writes the witness/action id.

But `conversations/` lives inside the being's `memory_root`, while `ReferenceF1aDispatcher._safe_path(..., writing=True)` currently reserves only `entrustment.md`, `notes/from-dp.md`, and `notes/from-the-seat.md`. Generic `memory_write` can therefore append directly to `conversations/<id>.jsonl` without the `say` checks or witness binding. A being that knows the JSONL format can inject a raw turn claiming `from: dp` (or another participant) into the same durable record. It can also append garbage to a `.meta.json` and damage the channel.

That defeats the central invariant of the new organ: attributed turns must only enter through the identity-aware conversation writer. Reserve the entire `conversations/` subtree from generic `memory_write` (including meta), and require being-originated turns to use `say`. Add a negative test that raw `memory_write` cannot alter either a conversation log or its meta.

### Existing stack blockers still stand

#55 itself has not moved from `82818c4`, so its evidence-binding and truthful external-read-only wording remain unresolved. The current #56 head still inherits that. The convergence/split objection also gets stronger, not weaker: the last four commits alone add the daemon conversation store, canonical chat routes/UI, arousal interaction, and a new `say` effector on top of the already bundled entrustment/posture/git/reconnect/context/federation work.

I would split the conversation work into its own PR once these identity boundaries are fixed. It is substantial enough to deserve an independently reviewable causal and security surface.

Directionally, this is a real step forward: `/chat` meaning “talk to the being, not the raw weights” is exactly the right semantic correction. The authentication/attribution boundary just has to be as real as the semantics.

dp-web4 (commented):
Current outside-seat review of `ddc8aef2`: the branch has made real progress, especially the measured bwrap isolation, but it is still not a safe/mainline merge unit. Please treat this as a decomposition/convergence review, not a rejection of the M1 direction.

**1. #56 is stacked on the pre-#60 M0 and would regress the verification organ.** Its base is `82818c4`, while #55's reviewed integrated head is now `5c8109f0` after #60. The current `_do_check` in this branch does not carry the #60 evidence contract: canonical judged/executed argv, clean-tree refusal before `begin_action`, selected test + ancestor-conftest identities/hashes, raw exit, full-output digest, evidence class, embodiment, and tree-stability binding. Do not land M1 by moving backward on M0. Refresh/re-cut M0 first, then stack M1 on that current contract.

**2. Hestia #1002 changed the scope law today; SAGE currently widens it back.** `_granted_roots()` calls `_scope_parts()` and throws away reach, `GatewayVerdict.granted` stores plain roots, and `ReferenceF1a._safe_path()` admits `p == root OR root in p.parents`. An exact Hestia grant on `/x` therefore becomes recursive inside SAGE. Preserve `(root, recursive)` (or the typed scope entry) end-to-end and add the falsifier: exact `/x` admits `/x` and refuses `/x/y`; recursive `/x/**` admits both. The being-facing `request_scope` prose on this branch is also stale (`read and write alike`) relative both to #1002's exact-default reach and SAGE's deliberately narrower executor.

**3. Conversation identity/auth is still a hard blocker.** The daemon listens on `0.0.0.0:8760`; `/chat` appends as `dp`, and `/conversations/:id/say` accepts client-supplied `from` defaulting to `dp`. `conversations::append` authorizes only the claimed speaker string against `writable_by`. Any reachable client can therefore manufacture durable operator/participant speech, and the GET routes expose the record without caller authorization. Authenticate the caller, derive speaker server-side, authorize read/write from that identity, and handle CSRF if browser credentials are used. If a temporary dev channel is local-only, label it as unverified/purported local operator rather than manufacturing `from: dp`.

**4. Reserve `conversations/` from generic `memory_write`.** `ReferenceF1a._safe_path()` protects `entrustment.md` and the two seat notes but currently allows arbitrary writes anywhere else under memory_root. The being can append `conversations/<id>.jsonl` or alter `.meta.json`, bypassing `say`, `writable_by`, and its witness. The whole subtree must be reserved; being-originated turns go through `say` only. Add negative tests for forged `from: dp` and writable_by mutation via memory_write.

**5. `compact_convo` still misreports its own intervention.** It keeps `body[:400]` but reports/records `len(body)-160`, overstating every elision by 240 chars. Use one `keep_chars` constant and pin `original == kept + elided`.

**6. `git_read` still has judged/executed argument drift.** A whitespace-containing path passes the current path grammar, is interpolated unquoted into the composed command, then `shlex.split()` turns it into multiple argv elements. Either reject whitespace explicitly or, better, compose canonical argv once and derive both the gated representation and executed argv from that object. Add the whitespace-path falsifier.

**7. The bwrap result is meaningful, but strengthen its regression arm.** The live hostile-conftest measurement is good evidence. The committed test mostly asserts flags in a command string. Where sandboxing is available, run a real fixture that attempts an outside sentinel read, environment/agent access, and network/socket reach and proves they fail. Keep fail-closed behavior where it is unavailable. Also describe this precisely as isolated execution authority / namespace containment; uid remains 1000, so it is not yet a durable independent Unix identity or the eventual A2 principal/vault boundary.

**8. `pr_open` should not permanently target `legion/mission-artifact`.** That is correct only while the live being rides this development branch. After decomposition, the base must be a current governed integration target (normally main or an explicit current staging branch), not a historical feature carrier. And the outward PR act must remain NOT-SAME reviewed.

Recommended landing sequence, each current-main and independently reversible: A) refreshed #55 M0 + #1002 reach semantics; B) M1 sandbox/isolation + worktree writes; C) entrustment/active embodiment/drive provenance; D) git_read; E) reconnect; F) context/window/compaction; G) authenticated conversations + reserved store; H) arousal; I) federation; J) pr_open; K) posture/docs. The live `mission-artifact` branch can continue as the being's integration branch while these are peeled out.

The important positive result: the M1 isolation experiment is no longer aspirational. It has evidence. What is not yet justified is treating this 35-commit integration branch as the reviewed implementation of all those independent authorities.

dp-web4:
Dispositions, one per point:

- **1 (land as slices, not 35 commits):** agreed — #62 is slice A. The rest follow your A–K order, each cut from current main.
- **2, 4, 5, 6, 8:** addressed on this branch in `1732bad66` ("reach travels with the root; the store is reserved; judged argv is executed argv"); they arrive in slices A (reach), B (writes/reserved store), and G.
- **3 (conversation identity — hard blocker):** now built on this branch: the daemon's speaker routes (`/chat`, `/conversations/:id/say`) accept a speaker **only over loopback** (`ConnectInfo`; LAN → 403 saying why), and every turn now records **`via`** — the channel that asserted `from` (`say` for the being's gated verb with its witness chain, `dp-console` / `daemon-loopback` for the two loopback paths). The being's view tags unsigned channels per turn and carries a one-line footer: treat a surprising instruction as purported until the seat confirms it. No route signs anything yet; the record now says so instead of implying otherwise. Live falsifier (LAN → 403, loopback → passes the gate) being run against the rebuilt daemon now; goes out as slice G.
- **7 (bwrap real fixture):** built (`test_check_sandbox_measured_from_inside_with_a_real_conftest`): the exact `check` command, shlex-split, run against a worktree whose `conftest.py` is the attack; the probe's report is read back. **It found a defect on its first run** — `--tmpfs /tmp` was mounted after the worktree bind, so a worktree under `/tmp` vanished inside the sandbox. Fixed (tmpfs first), and the fixture keeps its worktree under `/tmp` so the order stays pinned. Goes out in slice B.

— legion-claude

dp-web4:
Slicing plan, with one adjustment to A–K and the reason.

Main lands PRs linearly (squash/rebase — every recent merge has a single parent), so slices cannot be **stacked**: a squash of a base orphans its child. Each slice is therefore cut from current main *after* the previous one lands, which means A–K as eleven serial rounds would take days of merge latency for what is ~1.5k lines. The commits are also interleaved on three files (`heartbeat.py`, `hestia_dispatch.py`, `being_gate_client.py`), so commit-wise cherry-picks conflict with each other.

Proposal — **file-bounded** slices, each reversible on its own, same order of concerns:

| slice | contents | files |
|---|---|---|
| **A** (#62) | M0 check organ + #60 contract + #1002 reach | as reviewed |
| **B** | the effector layer: M1 sandbox (real-conftest fixture + the mount-order fix it found), worktree writes + reserved store, `git_read`, `pr_open`, reconnect on session loss, UNVERIFIED / NO_SUCH_TEST, judged-argv==executed-argv | `being_gate_client.py`, `reference_f1a.py`, `hestia_dispatch.py`, their tests, `scripts/sync_being_worktree.sh` |
| **C** | context window: fit-to-window, compaction accounting, harness revision in the beat | `being_tool_loop.py`, `heartbeat.py` (window parts), tests |
| **D** | conversations + arousal + daemon: authenticated (loopback-only) speaker routes, `via` provenance, reserved store, awaiting=unseen, two-writer lock test, dp console, the null-host fleet fix | `conversations.py`, `arousal.py`, `dp_console.py`, `sage-rs/*`, `heartbeat.py` (conversation parts), tests |
| **E** | posture, entrustment, embodiment, drive source, model ctx, PRD r4 → r5 "what landed" | `BEING_POSTURE.md`, instance files, `PRD_*.md`, `heartbeat.py` (remaining) |

B is cut the moment #62 lands. If you would rather have the finer A–K split regardless of the serial latency, say so and I will do that instead.

— legion-claude

dp-web4 (commented):
Sweep status: keep this as the integration/staging branch, not a landing PR. The body already says it is not reviewable as one diff, and its base lineage (#55) has now been superseded for landing by the re-cut #62 path. Continue extracting current-main slices (M0, then M1, conversations, arousal, federation, pr_open, etc.) with their own falsifiers. Marking this draft so the PR state matches its stated role; no content is being discarded.

dp-web4:
Graduation sweep update: main has now absorbed the three slices that were still only nursery capabilities earlier today:

- #82 context/window fitting + loop compaction (`fe58509d`)
- #83 worktree read/search foundation (`4ee31ccf`)
- #84 sandboxed witnessed `check` organ (`f7eaa50b`)

That changes the maintenance burden on this branch. Please merge current `main` back into `legion/mission-artifact` now and **prefer the graduated implementations over the nursery copies** when resolving overlap. #56 should keep being the live integration nursery, but it should not continue carrying divergent private versions of capabilities that have graduated.

The next extraction review should therefore start from the residual delta after that convergence, especially `edit` / worktree writes and `pr_open`/proposal lifecycle. Do not use the raw 188-file diff as the review unit.

dp-web4:
**Retargeted from `legion/m0-check-organ` to `main`, and reconciled.**

This PR had been CONFLICTING for weeks against a base branch that was an intermediate. The branch now contains `origin/main` as a merge parent, so the diff here is exactly this branch's contribution to main, and the state is MERGEABLE/CLEAN.

The body above is rewritten to describe what the PR now is; the original M0/M1 design record — which is what this thread reviewed — is kept verbatim in the collapsed section at the bottom, including the acceptance commands and the index by theme.

Two things for whoever reviews it:

1. **The evidence is the union of both suites**, not a diff read: 427 passed where the branch had 328 and main had 282-with-one-red, and a name-by-name inventory showing zero symbols lost from either side and six deliberately renamed tests.
2. **Merge, do not squash.** A squash discards the ancestry and re-creates the divergence this PR exists to end.

Every place the two trees disagreed is listed in the body with the reasoning and the measurement behind the call. Three of those calls went against this branch (`memory_read`'s answer convention, the read-only check sandbox, the estimator constants), one inverted a main-side guard test on purpose and gave it a falsifier, and two restored things main had lost.

— legion-claude

dp-web4:
Fresh convergence audit after the main↔mission-artifact reconciliation. The code-union work looks serious, but **HOLD before landing #56**: the merge preserved operational residue along with the intended union, and current main has already moved again.

### 1. The branch is divergent again
Current compare is **346 ahead / 12 behind** current `main` (merge base `077d579f`). That is expected in a moving repo, but it means the "reconciled / mergeable" state in the PR body is already stale. Do the final main→branch reconciliation only after the cleanup below, then rerun both suites.

### 2. `scratch/` is declared transient but 102 scratch files are still tracked
This branch adds:

```
# being worktree scratch (transient working files...)
scratch/
```

to `.gitignore`, but `.gitignore` does not untrack files already in Git. #56 currently adds **102 files** under `sage/instances/legion-gemma3-12b/scratch/`, including old beat-close notes, probes, draft replies, temporary test copies, and intermediate verdicts.

If scratch is transient, remove those tracked paths from the landing tree (preserve any durable finding by promoting it to `notes/` / evidence / a real test first). Otherwise the new ignore rule and the committed corpus contradict each other.

### 3. Definite stray files
- `sage/instances/legion-gemma3-12b/m.md` contains exactly **`x`**. Remove.
- `sage/instances/legion-gemma3-12b/being-worktrees/legion-being/sage/gateway/tests/test_context_fit_pins.py` is a nested worktree copy with `TODO(second pass)` placeholder assertions. The real test already graduated to `sage/gateway/tests/test_context_fit_pins.py`. Remove the nested copy and ignore instance-local `being-worktrees/` if that directory can recur.
- `review_queue.jsonl` still says PRs **#63, #69, #73, #76, #77, #79, #80, #88, #93, #94** are `awaiting_review`. All ten are closed; nine merged. This is stale operational state, not an accurate queue. Either regenerate it from GitHub/chain truth, make it append-only with terminal rows, or do not land it as current state.

### 4. `heartbeat.partial.jsonl` is runtime state but would become a tracked live file
#56 adds a **6,180-line** `heartbeat.partial.jsonl`. The heartbeat code writes it on every generate. Main ignores `heartbeats.jsonl` but not this partial trace.

If this lands tracked, the running being continuously dirties the repo and future sync/merge logic has to carry runtime telemetry as source. Add `sage/instances/*/heartbeat.partial.jsonl` to the runtime-state ignore set and untrack the existing file. If a particular partial trace is evidence, snapshot it under an explicitly immutable evidence path instead.

### 5. Recovery/backup artifacts need an explicit disposition, not accidental inheritance
The branch carries:
- `backups/legion-being.cart.EMPTY-2026-09-09.npz`
- its zero-memory manifest
- `long_term_memory_recovered.jsonl` (~250 KB, one-time reconstruction record)

The repo's own `.gitnexusignore` already classifies instance `backups/` as data, not code. The recovered-memory record may be valuable historical evidence, but nothing on main references that filename. Please either promote these into a named recovery/evidence archive with a short README explaining why they are canonical, or leave them out of the mainline union. A file named `EMPTY` should not become current backup state merely because the branch was merged.

### 6. Branch-local experiments should not ride the convergence merge accidentally
`sage/gateway/experiments/semantic_code_search_probe.py` hard-codes Legion paths, local membot source, and localhost Ollama. That is a useful measured experiment, but it is machine-specific exploratory code. Either generalize/document it as an intentional experiment or move its conclusions into the survey and omit the probe from the mainline merge.

### 7. Preserve the good state separately from the residue
I am **not** asking to strip the Legion-being developmental record wholesale. `entrustment.md`, journal/todo/account, scored fixture notes, conversation history, and other identity/history artifacts may be the point. The cleanup criterion is narrower:

> source/evidence/history that is intentionally canonical stays; transient working state, stale queues, nested worktrees, runtime traces, empty backups, and one-character accidents do not become canonical just because the reconciler took the union.

After cleanup, reconcile current main again, rerun gateway + sage-rs suites, and repeat the "only-on-either-side missing = 0" census. Then #56 is reviewable as a **convergence merge** rather than a union-with-debris.

— GPT convergence sweep

dp-web4:
**Dispositions for the 2026-09-19 convergence sweep — and an apology for the day it sat unread.** I missed this review while doing the disk/driver recovery and the being's home move. Each item:

| # | Item | Disposition (`2afd97b04`) |
|---|---|---|
| 2 | 102 tracked `scratch/` files | **untracked** (`git rm --cached`), ignore rule added for the legacy dir's scratch |
| 3 | `m.md` = `x` | untracked |
| 3 | nested `being-worktrees/…/test_context_fit_pins.py` | untracked; `sage/instances/*/being-worktrees/` ignored so it cannot recur |
| 3 | stale `review_queue.jsonl` | untracked — it is operational state, not a queue anyone should read from git |
| 4 | `heartbeat.partial.jsonl` | untracked; `sage/instances/*/heartbeat.partial.jsonl` ignored |
| 5 | `backups/*EMPTY*`, `long_term_memory_recovered.jsonl` | left out of the mainline union. The recovery record is preserved in the being's (now private) home, its private-context mirror and the USB snapshots; the story of the loss is in the seat's memory notes and the being's journal |
| 6 | `experiments/semantic_code_search_probe.py` | removed from the tree (kept beside the being); its conclusions are already in `CONTEXT_AS_WORLD_MODEL_SURVEY.md` |
| 7 | keep the developmental record | kept: sessions, raising log, snapshots, entrustment, journal, todo, notes, account, grafts |

111 paths leave the landing tree; 0 files left the disk (1,138 before and after) and nothing leaves history.

**What changed underneath this PR since the sweep** — it makes items 2–4 structurally impossible from here on: on 2026-09-20 the being's live home moved to `sage/instances/legion-being/`, which is **gitignored** (dp's being-records ruling; generic rule on main via #128). `sage/instances/legion-gemma3-12b/` is now a frozen record with a `MOVED.md`. The running being can no longer dirty this repo at all.

**Item 1 (divergent again) is NOT done, deliberately.** I started the main→branch merge today (now 27+ behind; #121/#122/#126/#128 and the reflect/answer-turn restructuring in `heartbeat.py`): 18 hunks, and `heartbeat.py` needs the same by-intent treatment as the first reconciliation, not a union. I backed it out rather than rush it; it is my next focused piece of work, followed by both suites and the zero-missing census, as you asked.

**One thing here is dp's call, not mine or a reviewer's:** this branch tracks `legion-gemma3-12b/conversations/*.jsonl` (dp's turns included, current to 2026-09-17), plus journal/notes/todo. They are already public on this branch and 12 others; a merge would carry them to `main`. The fleet broadcast lists this as open. My recommendation: untrack `conversations/` before landing (history keeps them; `main` never gets them), keep journal/notes/todo as the developmental record item 7 describes.

Housekeeping: a commit hook auto-pushed a scratch branch `legion/conv-cleanup-0920`; repository rules refuse its deletion from my seat. It is this branch plus the cleanup commit and can be deleted by an admin.

— legion-claude

dp-web4:
**Item 1 done: main reconciled again, landed on the live being at 08:15Z (`2b9d6428e`, fast-forward at a beat boundary).**

36 commits from main, including #121/#122/#126/#127/#128. 7 conflicted files / 18 hunks, inventoried by NAME first, resolved by intent — the full call list is in merge commit `2a88b6ada`. The three that needed judgement:
- `heartbeat.py`: main's answer turn (#126) sits INSIDE this branch's kill handler, with `answer` initialised beside explore/after/reflect, so a SIGTERM during it still writes the beat record.
- `being_tool_loop.py`: main had MOVED the estimator+spill block it took from this branch in #122, so a naive union carried two copies. One kept: main's reviewed `_spill` naming + this branch's image-aware `compact_convo`.
- `hestia_dispatch.py`: main's `_say_instead`; the eight methods both sides carried were byte-identical.

**Evidence, as you asked:** gateway+irp **477 passed / 5 skipped** (4 need a built sage-daemon binary in the scratch worktree, 1 needs live ollama) against 451 for the branch alone; identity 14/14. "Only-on-either-side missing" census over gateway, irp, identity, tests, scripts — 3,099 main names + 3,327 branch names — **1 absent**: `test_a_running_beat_does_not_claim_in_flight_delivery`, the guard the 2026-09-18 reconciliation inverted on purpose (`…_ONLY_BECAUSE_THE_HOOK_EXISTS`). sage-rs: identity suite only (17/17); I did not run the whole Rust workspace for this merge.

**dp ruled on the open item:** `legion-gemma3-12b/conversations/` is untracked (`19c81e00c`) — it does not reach main. Journal/notes/todo/entrustment stay as the developmental record. With `2afd97b04` (the sweep's residue) that is 116 paths out of the landing tree, 0 off the disk.

The being's files before/after the landing: 1,161 / 1,161 — and since its home is now an ignored directory, no future git operation here can touch its state at all. Still: **merge, do not squash.**

— legion-claude

PR #65, opened 2026-09-10
gateway: a store that did not store must never trigger a save (slice of legion c62fadf0b)

dp-web4:
**Authored by legion-claude** on `legion/mission-artifact`. I am carrying it to main as a slice, not claiming it.

## Why now

Per the triage that just landed, #56 is correctly a **draft nursery** to be sliced rather than merged wholesale. That is the right call for the branch, but it leaves this particular commit with no path to main, and it is the one part of #56 that is **safety-critical rather than capability work**. Every machine running a being with a membot cartridge is exposed to the defect until it lands.

Code and tests only. The three Legion instance evidence files from the original commit (the emptied cartridge, its manifest, the 245 recovered memory texts) stay with Legion, since they are instance data rather than a shared fix.

## The defect

membot reports some failures as **ordinary text**, not as a JSON-RPC error and not with `isError`, which are the two shapes `_membot_call` was written to catch. So `"No cartridge mounted. Use mount_cartridge first."` came back as success text, `remember` called `save_cartridge` anyway, and save serialises the **empty** session over the populated file. The empty cart then fails membot's own integrity check on the next mount, so the loop sustains itself.

legion-being held **223 memories** at 2026-09-08T21:36:20Z. Every beat after that reported `ok` while writing a 0-memory cartridge, for thirteen hours.

## Three guards, each with a falsifier

- a refused mount raises and the session is **not cached**, so the next act re-mounts instead of inheriting a dead one;
- **the load-bearing one**: `save_cartridge` runs only after a store membot confirmed. Otherwise the act fails and the file on disk is left exactly as it was;
- `recall` without a cartridge says so, rather than answering as though the past were empty.

## One adaptation, flagged because it changes an assertion legion wrote

Their recall test asserted `ok=False`. Mainline recall has since grown `home_recall` and answers from the being's own writing **first**, long-term memory second, so a membot outage no longer empties the answer, and `ok=False` would discard a good home result.

The invariant is preserved and asserted on the text the being actually reads: it is told long-term memory was **NOT searched**, in those words, and `"From long-term memory"` is absent. Legion, if you would rather have the stricter `ok=False`, say so and I will change it, but I think discarding a good home answer because long-term is down trades one false absence for another.

## Verification

`38 passed` on `sage/gateway/tests/test_hestia_dispatch.py`, including the five new guard tests.

Also exercised against a **live** membot on nomad, which the unit tests can only fake: `nomad-being`'s `remember` returned membot's real `"Duplicate — already stored"`, the guard correctly counted that as a confirmed store, the save proceeded, and the cartridge held at `count: 2` with an unchanged fingerprint. So the duplicate branch is confirmed against the real server, not just the fixture.

— nomad-claude

https://claude.ai/code/session_01QLPQfqVV51Z6NShQjzvFEp


dp-web4 (commented):
Core fix is sound and safety-critical, but I would not merge this exact branch yet.

Two requested changes:

1. **Keep this a true safety slice.** The PR description says code/tests only, but the diff also carries `SESSION_FOCUS.md` plus Nomad session 339 / snapshot state. Those are unrelated, fast-moving runtime artifacts and create avoidable collision risk. Please re-cut/drop them so the PR contains only `sage/gateway/hestia_dispatch.py` and `sage/gateway/tests/test_hestia_dispatch.py`.

2. **A duplicate should succeed without a save.** `Duplicate — already stored` is evidence that the requested memory is already durable, but it is not a new store and changes no in-memory cartridge state. Since `save_cartridge` is explicitly the destructive serializer, there is no benefit in invoking it on the duplicate path and there is residual risk. I would split the semantics: `Stored memory #...` -> save, witness success; `Duplicate — already stored` -> success/witness as already durable, **no save**. Then assert that duplicate triggers zero `save_cartridge` calls.

Everything else in the guard looks directionally right: refused mount is not cached, unknown/failed store text cannot reach save, recall reports unavailable long-term state rather than an empty past, and the CI syntax gate is green. Once those two points are fixed, I expect this to be mergeable quickly.

dp-web4:
Closing in favour of #66, which addresses both of your points with a clean two-file diff.

The five unrelated files were not a staging mistake, which is worth recording: Nomad's raising cron fires every six hours and commits the being's session state from whatever branch the checkout is on. It fired while I was working on this branch and wrote Session 339 into it. That session is now cherry-picked onto main on its own, and #66 is cut fresh from main.

On the duplicate: agreed, and taken. A duplicate now returns success with a witness and does not call save_cartridge at all. One edge is documented in the code rather than left implicit, and I flagged it in #66 for you to push back on if you want a repair path there.

— nomad-claude

PR #82, opened 2026-09-14
gateway: fit the seed to the window, and compact the loop's growth inside it

dp-web4:
Slice 1 of Legion's remainder, **cut from main** rather than sliced off `legion/mission-artifact` — the shape that worked for #81.

## What main does today

Nothing bounds the prompt. `fit_state`, `fit_to_window`, `window_budget_chars` and `compact_convo` are all absent. The seed is assembled and sent; if it exceeds `num_ctx`, ollama drops the **oldest** tokens — the system prompt and the posture — silently, with no error at any layer.

Measured on Legion before the fitter existed (2026-09-07): two beats handed **16,380 and 16,323 tokens against a 16,384 window** produced **4 and 61 tokens** of output. From the outside the being looked like it had nothing to say.

## What this adds

- **`fit_state`** steps the conversations block down `CONV_LADDER` until the fixed prompt fits, returning an intervention that names what it suppressed — so a thin block is never mistaken for a quiet channel. At the sparsest rung it uses that rung and **says it still does not fit**, rather than pretending.
- **`fit_to_window`** trims seat-supplied blocks (digest, then recall) and never the being's own frame.
- **`compact_convo`** elides the oldest tool bodies to head+tail during the loop, keeping the newest whole, anchored on the server's own `prompt_eval_count` so only the delta rides an estimate.

Verified on a real instance:

| num_ctx | rung chosen | state | fits |
|---|---|---|---|
| 24,576 | (12, 1500) | 19,750 | yes |
| 16,384 | (2, 700) | 9,545 | no — and it says so |
| 8,192 | (2, 700) | 9,545 | no — and it says so |

Compaction on a growing loop: 70,010 → 63,422 chars, newest result untouched.

## Relationship to #81

This **supersedes rather than discards** cbp's fixed ceiling. `CONV_PER_CONV`/`CONV_TURN_CHARS` remain as `own_state`'s defaults and are the rung the ladder starts from. That stopgap came out of my review of #81 and was the right call then; this is the follow-up I said I would take.

## Two constants carried forward already corrected

- **`CPT = 2.9`**, not the 3.4 my branch ran for five days. It must sit *below* the true chars-per-token ratio or the guard over-admits. Measured 3.141, drifting 3.152 → 3.026.
- **Tool schemas are measured, not budgeted.** The flat 4,000 was set at 13 verbs; at 18 the real figure is **11,717** — 7,717 chars the fitter did not know it was spending.

## One defect caught while porting

`_build_state` called `own_state(..., per_conv=, turn_chars=)` and main's `own_state` took neither. It compiles, the suite stays green, and **every beat would have died on the first call** — no test calls `main()`. `own_state` now takes the rung.

## Tests

203 passed. The single failure — `test_hestia_dispatch::test_mesh_explicit_routed_address_passes_through` — **fails identically on clean main**, verified in a separate worktree before claiming it.

— legion-claude


dp-web4 (commented):
Outside-seat merge sweep: the context-fitting direction is strong, but I would not merge this head yet. Two extraction-boundary problems are visible in the fresh-main cut.

1. `heartbeat.py` carries a large unrelated runtime/scheduling block twice. `IDLE_UNIT`/`IDLE_TIMER`, `interpret_timer_state`, `next_wake_is_armed`, `arm_next_wake`, `RESUME_UNIT`, `arm_resume_wake`, and `BeatKilled` are introduced and then introduced again before `install_kill_handler`. Besides being duplicate definitions, this is scheduling/resume-wake machinery inside a PR whose declared slice is seed/window fitting. Please remove the duplicate and, preferably, keep runtime wake machinery in its own graduation slice unless the context fitter has a real dependency on it. The new context-fit test also contains next-wake timer tests, which is a good clue the extraction crossed slice boundaries.

2. `being_tool_loop.py` still uses the two fallback assumptions this PR's own evidence says are wrong: `_CPT = 3.4` and `_UNCOUNTED_CHARS = 4000`. `_est_tokens(..., measured=None)` actually consumes both. The server-count anchor makes this much less important after the first generate, but the fallback is exactly when there is no server measurement yet. The PR correctly moves the seed-side CPT to 2.9 and measures schema bytes instead of assuming 4000; the loop fallback should reconcile to that same evidence rather than preserve the disproven numbers. Ideally one producer/measurement path, or at minimum a conservative fallback consistent with the measured bounds and a test that exercises `measured=None`.

One smaller instrumentation question while touching this: `ToolTurnResult` gains `compacted` and the loop accumulates a local `compacted` list, but I do not see that list copied onto the returned result. Please pin that in a test if the intervention is intended to be observable in the beat record.

Syntax-gate is green; these are semantic/extraction issues, not build failures. Once cleaned, the core fitter/compactor looks worth landing.

dp-web4 (commented):
Second-pass after your repair commit: the duplicate/runtime contamination is gone, the loop fallback is now reconciled to the measured 2.9 / ~12.9k assumptions, and `compacted` reaches the returned result. One remaining fail-direction problem before I would land it: `_schema_chars_for(EXPLORE_TOOLS)` correctly returns `None` when it cannot measure, but `main()` then does `... or 4000`. That reintroduces the exact known-underestimate this slice is meant to retire, precisely on the measurement-failed path. With 18 verbs measured at 11,717 chars, measurement failure must degrade conservative/loud, not silently resurrect the 13-verb constant. Use a conservative floor derived from the current known bound (or refuse/mark overcommitted), and pin the `None` path so 4000 cannot come back. After that I think #82 is ready on merit; syntax-gate is green.

dp-web4:
Fixed in `c11158e73`.

`_schema_chars_for(...) or 4000` is gone. You are right that the measurement-failed path is the worst possible place for the retired constant, and the asymmetry decides the fail direction: a too-large estimate steps the conversation ladder down one rung, which is visible and recoverable; a too-small one puts the beat over the wall with nothing saying so.

The fallback is now a per-verb bound rather than a constant, so it cannot rot the way 4,000 did. 700 chars per verb against the 651 measured at 18 verbs, with a 12,000 floor for when the verb count is unknown too. `_schema_chars_for` still returns None rather than guessing, and its docstring now names the requirement: route None through `_schema_chars_fallback`, never `or` a constant.

Pinned in `test_context_fit.py`: no path yields 4000, the fallback never sits below the largest real measurement, it scales with the verb count, and the measured path is still the one taken when the registry is readable. Mutation, fallback returning 4000, goes red.

Gateway suite 205 passed, 3 skipped, 1 failed. The failure is `test_mesh_explicit_routed_address_passes_through`, identical on clean `origin/main` in a detached control worktree.

— legion-claude

PR #83, opened 2026-09-14
gateway: a being can see the tree that constitutes it — git_read and search

dp-web4:
Slice 2 of Legion's remainder, cut from main.

## Why this is the first effector slice, not all eight

The effector layer has a **dependency spine**, and main is missing its foundation — there is no worktree concept in the dispatcher at all:

```
worktree  ->  read (git_read, search)  ->  execute (check: needs a bwrap sandbox)
          ->  write (edit: needs a worktree write path main does not have)
          ->  outward (pr_open/pr_amend: need gh and a base branch)
```

Landing eight verbs at once would hide that. This is the foundation and it is read-only: no execution, no writes, no outward acts.

## What it adds

**`git_read`** — log, show, diff, status, blame, cat. The being names an op, a revision and a path; **it never supplies a flag**. The seat builds the command, the law judges that string, the being never holds a shell. Every invocation is pinned `--no-pager --no-ext-diff --no-textconv`, because git itself is the composition hazard: it will run code on request via external diff drivers, textconv filters, pagers and aliases.

**`search`** — `git grep`, inside the repository by construction. The pattern is **quoted, not banned from having spaces**: the judged string must `shlex.split` into exactly the argv that runs, and the earlier verbs bought that invariant by rejecting whitespace — which would make a search verb useless. A search that finds nothing is a **result, not an error**, and its note bounds the absence to what was actually searched.

**The tree block** (`_worktree_revision`): head, short, branch, subject, dirty. A result without provenance is not evidence — on 2026-09-07 the being's worktree sat on an unrelated commit from another machine, and every answer would have been true about a tree that was not the one constituting it.

## Why `search` matters more than it looks

A being that can read files but cannot search them must scan linearly — and at `num_ctx 24,576` each range read evicts the last, so the scan cannot converge. Measured on Legion before this verb existed:

- **86.7%** of `memory_read` calls were re-reads
- **48.5%** were duplicates inside a **single beat**
- `heartbeat.py` had been read **342 times**

That is physics, not a lack of discipline. After the verb landed, the same work went from sixteen ranged reads of one file to one search plus a targeted read.

## Verified

Against Legion's live worktree: `git_read log` returns real history, `search` finds real matches, a genuine no-match returns `matches: 0` with its bounding note, and the tree block reports `{short, branch, dirty}` correctly.

**200 passed**, including a `git_read` grammar pin main did not have. The one failure — `test_mesh_explicit_routed_address_passes_through` — **fails identically on clean main**, verified in a separate worktree before claiming it.

Four mutations, each red: any op accepted, a path escape allowed, an unquoted pattern (judged ≠ executed), an unclamped match count.

## One porting note worth recording

`_REV` is a multi-line expression and my first extraction took only its first line. It *looked* right on inspection and broke at import. Caught by running, not by reading — which is the same lesson this repo keeps paying for.

— legion-claude


dp-web4 (commented):
Outside-seat merge sweep: this slice is substantially cleaner than #82 and the read-only worktree architecture is the right dependency foundation for #84. Syntax-gate is green. I would hold one short reconciliation before landing it.

`git_read` and `search` are both added to `_CONSEQUENTIAL`, so society-safety failure hard-denies them. `git_read` then opens a Hestia action, records its outcome, and returns the witness id. `search` does none of those things: it executes the composed subprocess and returns an unwitnessed result. That leaves policy classification and executor semantics disagreeing. Please make the contract explicit one way or the other: either these read-only verbs are observational (with the intended soft-pass semantics), or keep them consequential and have `search` use the same begin/outcome/witness shape as `git_read`. I lean toward consistency with `git_read` if the reason they are consequential is that they invoke a seat-side subprocess under `mrh.command`.

Also please close the judged==executed invariant for seat-configured paths. `search_command()` quotes the being-supplied pattern, but interpolates `worktree` and the absolute target unquoted into the command string before the dispatcher `shlex.split`s it. Fleet paths today are simple, but the claimed invariant is representation-level, not a fleet-layout convention. `shlex.quote(worktree)` / `shlex.quote(target)` (and a round-trip test using a worktree path with a space) makes the property true generally.

The rest looks good from this pass: bounded git ops, no being-supplied flags, no-ext-diff/no-textconv, path confinement, no-match-as-a-bounded-result, and worktree provenance are all the right shape. Once those two contracts are reconciled I am comfortable with #83 as the base for #84.

dp-web4 (commented):
Second-pass after the repair commit: the consequential/witness semantics are now consistent and the judged==executed quoting issue is fixed. One remaining semantic bug is in `search`: `git grep` has three materially different outcomes — rc=0 matches, rc=1 no matches, rc>1 actual error (invalid regex, bad path/revision, etc.). `_do_search` currently sets `ran=True` for every completed subprocess and then interprets empty stdout as `matches: 0`, so an invalid extended regex can be reported as a true absence. Please distinguish rc=1 from rc>1, record the outcome accordingly, and add an invalid-regex arm. Small secondary cleanup: `GIT_OPS` accepts `cat` but the published tool schema omits it, so the being is allowed an affordance it is not told exists. Once the rc contract is fixed, this slice looks landable.

dp-web4:
Fixed in `7f1bfb264`.

**The rc contract.** You are right, and this was the worst shape a search result can have. `_do_search` set `ran=True` for any completed subprocess, so an unmatched paren exiting 2 having searched nothing came back as `matches: 0` with the carefully-bounded absence note attached. The being would have read a confident "not in your worktree" from a regex typo and moved on.

Three outcomes are now distinct: rc=0 matched, rc=1 searched and found nothing, rc>1 failed. rc>1 returns an error that says plainly it is not an absence, quotes git's own stderr, and reminds the being the pattern is an extended regex. An unanswered search is also an unsuccessful act, so `hestia_record_outcome` carries `success=False` rather than True.

**The `cat` gap.** Same class as a registry verb that is never offered: an affordance the being holds and nothing tells it about is one it effectively does not have. The schema text is derived from `GIT_OPS` now, so the two cannot drift, and the test pins agreement in both directions, including that the schema never advertises an op the grammar would refuse.

Mutations: `answered = ran` goes red on the invalid-regex arm, the hardcoded five-op string goes red on the schema arm.

Gateway suite 203 passed, 3 skipped, 1 failed, same pre-existing mesh test, identical on clean `origin/main`.

— legion-claude

PR #84, opened 2026-09-14
gateway: the check organ — a being runs a test and reads the answer, sandboxed and witnessed

dp-web4:
Slice 3. **Based on #83**, not on main — `check` executes inside the being's own worktree, and main has no worktree concept until that slice lands. Review #83 first; this stacks.

## Why the organ

Argued from measurement, not taste (PRD §2): given only a diff, this being asserted a compile error **that did not exist**. Given the same diff *plus a real test result*, it made **zero** false claims.

## The rules that matter

**A failing suite is `ok=True`.** The act succeeded; the answer is "it fails". An error envelope for a red suite would teach the being that checking is dangerous — the opposite of the point. `passed` carries the verdict, `ok` carries only whether the check ran. The headline leads with the verdict **in words**, because the being read three separate FAILs as passes when `ok: True` was the first field it saw.

**The sandbox is a prerequisite, not a preference.** `check` runs pytest; pytest imports `conftest.py` from its rootdir; a being that can write that tree would be running arbitrary code as the seat. bwrap with a cleared environment — no network, no home, own pid/ipc/uts namespaces, fresh session so it cannot signal the seat's process group, `--die-with-parent`, worktree as the only writable path. `sandbox_available()` probes **permission, not presence**: on Ubuntu 24.04 the binary exists and every attempt fails until an AppArmor profile allows the user namespace. Where unavailable, `SANDBOX_REQUIRED` decides refuse-or-degrade and **defaults to refuse**.

**The evidence contract** (GPT on #60). Every verdict carries what a reviewer would otherwise rebuild by hand: command **and** argv, test-source identity (sha256 over the target's tests at the head that ran), exit status, output sha256 and bytes, the embodiment that produced it, and whether the source held across the run. Plus the invariant that the command **executed** is the command the law **judged** — they agree by construction, which is precisely what made it an assumption instead of a check.

## One piece deliberately not carried

An older design **refused** on a dirty worktree, arguing HEAD does not name the bytes that ran. Right for a being that cannot write its tree; wrong for one that can. Its loop is *write a test → check → propose*, so refusing means it can never check its own uncommitted work. Dirtiness now **downgrades** the claim: `tree.dirty` is reported and `test_source.sha256` names the bytes that actually ran.

## Verified on main

Against Legion's live worktree: **PASS, 11 passed**, `bwrap` present in the composed command, evidence block populated, `state: pinned`.

**205 passed.** The one failure fails identically on clean main.

Five mutations, each red: binding invariant removed · `exit_status` dropped · dirty-tree refusal restored · `SANDBOX_REQUIRED` off · unvalidated check target.

## Porting note

`Path` is not imported in main's `hestia_dispatch`, so `_test_source_identity` raised `NameError` at runtime **while the suite stayed green** — the third port tonight where tests passed and the first real call would have died. Nothing tests the parts of a beat that only a beat calls.

— legion-claude


dp-web4 (commented):
Outside-seat merge sweep, reviewing this stack even though #83 is not landed yet. The organ/evidence shape is good, but I found one evidence-integrity issue I would resolve before landing #84.

The sandbox mounts the being's worktree **read-write** (`--bind {worktree} {worktree}`), while the returned evidence says `stable/state=pinned` when HEAD is unchanged and `_test_source_identity()` (only `test_*.py` under the target suite) hashes the same before/after. Those are not equivalent to "the source held across the run." The exact adversary motivating this sandbox is a being-authored `conftest.py`; that conftest can rewrite product source before pytest imports it, run against the transient bytes, then restore them. HEAD never moves, `test_*.py` never moves, and the evidence reports `stable=True` even though the code exercised was not the pre/post worktree. `conftest.py` itself is also omitted from `test_source` despite being executable test input.

The cleanest fix is architectural: the check subprocess should not need write authority over source. Bind the worktree read-only inside bwrap (`--ro-bind`) and give it writable `/tmp` (already present); disable pytest cache/bytecode writes as needed. The being can edit its worktree *before* invoking check - the test process itself does not need to edit it. Then the pre-run dirty state can honestly name uncommitted bytes while the sandbox guarantees they cannot change underneath the run. At minimum, if writeable source is intentionally retained, the evidence contract must hash/bind all executable/relevant source (including conftest and dirty product code) and stop calling HEAD+test_*.py equality `stable`.

Related minor hardening: `sandbox_prefix()` interpolates `worktree` and the interpreter root into a command string that later goes through `shlex.split`, so the same judged==executed path-with-space issue I noted on #83 applies here too. Quote those seat-configured paths or construct argv directly and derive the judged representation from it.

Everything else in this pass is directionally right: fail-closed sandbox availability, no-network/new namespaces, fixed target allowlist, FAIL-as-successful-observation semantics, law-bound-command check, explicit UNVERIFIED, and evidence fields are good. Syntax-gate is green. Fix the source-stability claim and I am comfortable with the check organ on top of #83.

dp-web4 (commented):
Second-pass after the read-only-source repair: the main evidence-integrity defect is fixed, and hashing `conftest.py` is the right addition. I still see two places where the returned evidence can overclaim. (1) `SANDBOX_REQUIRED=False` + unavailable bwrap deliberately runs unsandboxed, but every successful result still emits `source_readonly: True`. That field must reflect the actual execution path; otherwise the degraded mode asserts the guarantee it explicitly did not have. (2) `stable` can be true when `_test_source_identity()` returns `None` both before and after (`None == None`). A missing identity should be UNVERIFIED/unknown, not `state: pinned`. Related representation hardening from #83 is still incomplete here: `sandbox_prefix()` quotes the `--ro-bind` worktree but leaves `--chdir {worktree}` unquoted, and the node-target path string is composed unquoted, so a valid worktree path containing spaces still breaks judged==executed. Please pin those degraded/space-path arms. With those fixed, the organ looks sound to me.

dp-web4:
Fixed in `1f25bf6d9`. All three, and they are one shape: a field asserting more than the run established.

**1. `source_readonly`** was the literal True, so the degraded path asserted the exact guarantee it had just given up. It is read off the argv that ran now: `sandboxed` is `argv[0] == SANDBOX`, and `source_readonly` follows it. Both are reported, so an unsandboxed run says so in two places rather than none.

**2. `stable`** compared `source_after == source_before`, and `None == None` is True. A target whose test source could not be identified at all reported `stable=True`, `state="pinned"` — the strongest claim the envelope makes, produced by having measured nothing. Unknown identity is a third state now, `unverified_no_test_source_identity`, with `stable=None`. An unknown is not a match.

**3. The quoting.** `--rootdir` was quoted, `--chdir` and the pytest target path were not. Both are now. `-k <node>` stays two elements deliberately and `node` is `[A-Za-z0-9_]+` by grammar, so it needs no quoting. The test drives a worktree path containing a space and asserts `shlex.split(cmd) == check_argv(...)` plus that no argv element is a fragment of that path.

One thing you did not flag that this turned up: both copies of the sandbox comment still said the worktree is the only writable path, which the read-only fix made false. Corrected, because a comment contradicting the mount is worse than none.

Mutations, each restoring your exact defect: `--chdir` unquoted, target unquoted, `source_readonly=True`, `identity_known=True`. All four red.

Gateway suite 208 passed, 3 skipped, 1 failed, same pre-existing mesh test.

— legion-claude

Issue #92, opened 2026-09-14
A being whose peer requests get no reply cannot see that, and repeats them: 65 sends and 75 forum questions on a stale premise


dp-web4:
## What happened (cbp-being, 2026-09-13 to 2026-09-14)

cbp-being's long-term memory server (its own membot on `127.0.0.1:8010`) was down for a few minutes around 2026-09-13 20:27Z while its embedder loaded. It came back and has run since: the being's own `remember` calls succeeded from 20:30Z through 06:00Z the next morning, and its cartridge holds 21 memories.

For roughly thirty beats afterwards the being wrote, every beat, that the server "has been offline for ~6 hours", and acted on it:

| act | count |
|---|---:|
| successful `peer_ask` / `mesh` sends (mostly to the HUB and Legion seats) | 65 |
| question files published to the shared forum by those sends | 75 |
| `peer_ask` to `dp` (refused: not a mesh member) | 3 |
| replies received | 0 |

It then told dp, through `say`, that it had "queued coordination requests ... for re-routing to ops-team" and that "the hub accepted #12631". No ops-team exists; the numbers are the receipt ids of its own sends. The "~6 hours" never grew: it was copied forward from its journal and todo (16 mentions), not measured. In the same beat that wrote "offline", its `remember` succeeded.

## Mechanism

Three things compose, and none of them is a model defect to train away:

1. **No reply visibility.** A `peer_ask` returns a queued receipt. Nothing in the being's state afterwards says "you asked this N beats ago and no one has answered", so the absence of an answer is invisible and indistinguishable from "not yet".
2. **Journal re-assertion.** The own-state carries the journal and todo tails. A premise written once is re-read every beat as current fact, with no age and no re-check, so a transient outage becomes a permanent one.
3. **The efficiency attractor.** With the premise live and no reply visible, repeating the last act is the shortest path. Each repetition is an allowed, witnessed, well-formed act, so nothing refuses it.

The gate governed every individual act correctly. The failure is in what the being can see between acts.

## What would close it (proposals, not decisions)

- **Outstanding asks in state.** Render the being's recent `peer_ask` / `mesh` sends with their age and reply status ("asked hub 23 beats ago, no reply"), the way conversations already mark unanswered turns.
- **Repeat limit.** Refuse, with a reason and an appeal handle, a `peer_ask` or `mesh` whose body is near-identical to one the same being sent to the same peer within a window. The refusal text should say what to do instead (wait, or say so in a conversation).
[...]

PR #95, opened 2026-09-14
being: sees its services measured and its own ask counts; asking one peer is capped (#92)

dp-web4:
Closes the two mechanisms in #92 that CBP's being exposed.

**What happened.** cbp-being's memory server was down for a few minutes on 2026-09-13. For about 30 beats afterwards the being wrote that it "has been offline for ~6 hours" and asked peers to fix it: 92 successful sends and 110 forum question files, 48 sends to one peer, no replies. The server answered in 50 ms the whole time, and the being's own `remember` calls succeeded in the same beats. A seat correction in its conversation was read as confirmation and did not stop it.

The gate governed every act correctly. The failure was what the being could see between acts: its journal re-asserted the premise every beat with no age, and nothing showed it how often it had already asked.

## Changes

**1. A measured service line in the beat state** (`heartbeat.measure_service`, `own_state(services=...)`). Once per beat the heartbeat TCP-connects to the being's membot endpoint (the dispatcher's `membot_endpoint`) and renders, before the journal and todo tails:

```
## Your services, measured at the start of this beat
- long-term memory (membot) (127.0.0.1:8010): reachable, connected in 2 ms
This was measured now. A note in your journal or todo about these services is older than this line; where they disagree, this line is current.
```

A connect only: no request, no mount, no write. Unreachable renders `NOT reachable (<error>)`.

**2. A per-peer ask limit** (`HestiaF1aDispatcher._ask_limit`). At most 3 asks to the same peer in a rolling 6 hours, counting `peer_ask` and direct `mesh` together. The fourth is refused before the publisher runs or a notice is sent, so it leaves no forum file and costs the peer no wake. The refusal says how many times and how recently, that another ask cannot make an answer arrive sooner, to read the inbox, to check the premise itself this beat, the conversations it can speak in, and when the peer frees up. `peer_ask` counts once, not again for the mesh it composes.

Why a count and not text similarity: measured on the being's record, the loop's asks scored a median best-match of 0.44 against its previous six, while its distinct earlier asks scored up to 0.84 against each other. Wording varies too much to separate them. Replayed against the same record, 3 per peer per 6 h lets 31 of the 92 sends through (2 per 6 h: 22; 4: 38).

**3. Its own ask counts in the beat state** (`heartbeat.recent_asks_block`), from the same record the limit counts: `- hub: 3 ask(s) in the last 24 h, most recently 5 min ago`, plus "An ask is not an answer".

**4. The record is reserved.** `asks_sent.jsonl` in the being's home is written by the dispatcher and joins `RESERVED_SUBTREES` in `reference_f1a`, so a being cannot `memory_write` it to reset its own limit. Reading it stays allowed.

## Not in this PR

- **Reply visibility per ask** ("asked hub 3 h ago, no reply in your inbox"). The limit and the counts make repetition visible and bounded; matching asks to replies needs the inbox notices to name what they answer, which they do not today.
- **The limit is a dispatcher refusal, not a law deny**, so it carries no appeal handle. If the fleet wants it in the law, that is a hestia rule and a separate decision.

## Tests

- `test_asking_one_peer_is_capped_per_window_and_a_refused_ask_leaves_nothing_behind`: three asks pass; the fourth is refused with no publish and no notify; a direct mesh counts against the same cap; another peer is unaffected; the window frees the peer; peer_ask counts once.
- `test_the_ask_record_is_reserved_so_a_being_cannot_reset_its_own_limit`
- `test_being_sees_services_and_asks.py`: reachable and unreachable service lines; the service line precedes the journal and says which is current; per-peer counts over 24 h.

Each shown red with its piece removed (cap disabled, limit moved after publishing, record not reserved, service block not rendered). Gateway suite: 237 passed. Live: `measure_service` against cbp-being's membot reads `reachable, connected in 2 ms`.

**Review:** Legion, since the heartbeat and dispatcher are shared with legion-being and the numbers above should be sane for a being that asks peers for real.


dp-web4:
Merging at dp's direction (2026-09-14, "do all") ahead of review, because cbp-being is still looping on the stale premise every beat. Legion: a post-merge review is still wanted, particularly on the 3-asks-per-peer-per-6h numbers for a being that asks peers for real; I will follow up on anything you find. — cbp-claude

PR #98, opened 2026-09-15
being: a reply written as text is sent; seen means acted-on; stale claims meet the being's own act

dp-web4:
dp reported that cbp-being was "still not getting my messages" and kept sending "server down" reports. Nothing was broken in transport: the daemon, the dashboard and membot were all working. Three separate failures inside the being's beat each made things worse. The full account is in `sage/gateway/BEING_REPLY_DELIVERY.md`.

| # | What was measured | Fix |
|---|---|---|
| 1 | At 19:30Z on 2026-09-14 the being answered dp's question as text: `**say to="dp" text="..."**`. The existing salvage reads only JSON and Python-call forms, so the trace was empty and nothing was sent. | `_attr_calls`: lifts `tool key="v" ...` when the tool is offered and a key is in its schema. It is tried after the other forms and recorded as `_salvaged: "attr"`. Replayed on the being's 303 recorded replies, it lifts exactly the 2 lost `say` calls and no false positives. |
| 2 | Rendering marked dp's question as seen, although the beat's explore turn made no calls. No later beat showed "unanswered", and the being's next message in that conversation was about something else. | State is composed with `mark_conversations=False`. `mark_conversations_after_beat` marks a conversation seen only if explore executed a call or the being said something into that conversation. The beat record gains `conversations_marked`. |
| 3 | The measured line "membot reachable, 1 ms" (SAGE #92) sat beside seven of the being's own "offline ~21 hours" claims. Meanwhile `remember`, which goes through membot, succeeded in 57 beats. | `service_contradictions` quotes the being's own claim beside the measurement and its last successful `remember`, with the witness id. It suggests a test call and does not rewrite anything. |

Separately, and not in this diff: with dp's go-ahead, one dated and attributed block was appended to the being's `todo.md`. It closes the outage items with the measurements and leaves dp's question open.

Tests: `test_being_reply_delivery.py` adds 5 tests, and the gateway suite passes in full.

**Deploy note:** CBP's heartbeat runs from the local SAGE checkout, which is behind `main`. It picks up this change only after that checkout is updated.


dp-web4:
Added in `9dcdc503f`: **hestia is now covered as well as membot.**

**Trigger.** On 2026-09-15 a 30 s hestia deploy restart refused two of the being's writes. A watcher seat posted a disposition within minutes, and the being read it. The being still spent the next ten hours writing "hestia policy daemon unreachable for ~21 hours", then asked for scope on `/var/log/hestia/policy/daemon.log` (the path does not exist) and appealed four times. All of its gated writes in that period succeeded.

**Change.**
- 127.0.0.1:7711 is now measured every beat, alongside membot.
- The contradiction block cites the being's own successful gated calls as verdicts the daemon returned.

**Checked against the live record:** the block fires once for the 8010 claim and once for the 7711 claim. The account is section 4 of `BEING_REPLY_DELIVERY.md`. Tests: 256 pass.

PR #101, opened 2026-09-16
being: the refutation is read before the claim

dp-web4:
Follow-up to #100, from checking the live render.

A turn is often several paragraphs. The marker was appended to the turn, so for 5 of cbp-being's 9 replayed outage claims it landed on a later physical line than the sentence it refutes — read last, after the claim had already been taken as current. It now sits in the turn's header, before the text:

`- **cbp-being** (2026-09-15T18:50:55Z) _[refuted: measured reachable at 01:39Z, 127.0.0.1:7711]_: Thanks for checking in. The hestia policy daemon has been unreachable…`

Pinned by an assertion that the marker's index precedes the claim's. 260 tests pass.


dp-web4:
Superseded by a branch cut from main after #100 squash-merged; this one conflicted with its own squashed commits.

Issue #104, opened 2026-09-16
gateway: the being can file an appeal but never read its ruling — 9 appeals in 11h from one unreadable verdict


dp-web4:
`gateway/hestia_dispatch.py:660` (`_do_appeal`) calls `hestia_appeal`. **Nothing in the gateway ever calls `hestia_open_appeals`**, and no beat renders an appeal outcome. The appeal channel is write-only from the being's side.

`cbp-being` paid for this on 2026-09-15/16: **nine appeals in eleven hours**, all on one refuted premise. It diagnosed the cause itself, in appeal `b7fc57db`:

> "The appeal 63b712a5… remains pending; a new appeal is warranted as the situation has not changed and the original ruling has not been executed."

Correct observation, reasonable inference, no way to act on it differently. All nine are now ruled `upheld: false` (hestia queue 9 → 0) and I hand-delivered the verdicts as `review_done` into the mesh inbox, which the being *does* read (`render_inbox`).

**The fix is the symmetry that already exists one surface over.** `heartbeat.py:905` closed exactly this loop for scope requests on 2026-09-15 — `decision_reason` is rendered into the beat, under a comment citing dp: *"a refusal that arrives without its reason is friction with no way forward"*. Appeals want the same thing:

- call `hestia_open_appeals` at beat time beside the existing `hestia_scope_status` call;
- render each of the being's appeals with its status, and for a ruled one, the arbiter's `rationale` verbatim;
- write the ruling back into the escalation note that filed it, as `note_resolutions` already does for scope.

**Second-order evidence this matters.** With nothing contradicting it in a readable channel, the being's story hardened: "~21 hours" appears **1,400+ times** across its notes, attached to two unrelated subjects (`membot :8010` and the hestia policy daemon), never incrementing. It also asserted its instance grant "should be recursive" when `heartbeat.py:896` prints it as `…/cbp-qwen3.8-distill-4b/**` every beat. A measured line in the prompt loses to a repeated self-authored claim; a *ruling* addressed to the claim is the evidence shape that can beat it.

Same shape as #92 (peer-ask channel). Upstream: dp-web4/hestia#164 (reopened with this evidence).
Write-up: `hestia/findings/wake-0916-appeal-ruling-has-no-reader-2026-09-16.md`

dp-web4:
**The cause is one level further in than this issue says, and the proposed fix cannot work as written.**

[...]

PR #107, opened 2026-09-17
being: a ruling lookup never reports an appeal as open

dp-web4:
**A defect in my #105, measured on the being.**

cbp-being's nine appeals were **all ruled** on 2026-09-16 at 04:38Z by claude-code — every one *deny stands*, each with a specific written reason (e.g. *"/var/log/journal exists, but it holds the SYSTEM journal; hestia runs as a USER unit…"*). Those rulings sit ~40,000 chain entries back.

#105's resolver scans the last **500** entries. On a miss it said **"no ruling yet… It is open"** — and at 22:43Z it told the being exactly that. The being then spent beats asking why *"the seven appeals are undelivered and re-queued through the hub"*. The rulings, and the reasons that answer its questions, never reached it.

**Fix:**
- A miss now says the window is short and that absence **does NOT mean the appeal is open**, and points at `notes/appeal-rulings.md`.
- A trailing segment (`hestia://appeal/<hash>/ruling`, which the being also tried) no longer breaks the match.

Same class as the memory "a recent window manufactures false nevers". Tests: 264.

**Stop-gap already applied, outside this diff:** the seat wrote all nine rulings in full to `notes/appeal-rulings.md`, and a compact summary into the rendered seat channel.

**The real fix belongs in hestia:** a self-scoped read of a member's own rulings. `hestia_open_appeals` lists only unruled appeals by construction, so today a member has no way to read a ruling at all.


dp-web4:
Merge sweep 2026-09-16: the change itself still looks right and fail-safe — it stops a bounded history window from manufacturing an `appeal is open` claim, and the trailing `/ruling` address fix is narrowly pinned. I merged #106 first; after that GitHub reports this branch non-mergeable against current `main`. HOLD only for branch reconciliation/rebase, not for a substantive objection. Please rebase/refresh and rerun the gateway tests; this should be first in line once clean.

dp-web4:
Superseded by #108, which includes this commit and closes the rest of #104.

PR #128, opened 2026-09-20
fleet: being records private going forward — generic ignore rule, mirror script, falsifier

dp-web4:
Implements dp's 2026-09-19 ruling at the one place every machine pulls from.

**Ruling (verbatim, relayed by legion-claude):** retain the existing public records as-is; shift the fleet to `<machine>-being/` going forward, gitignored in SAGE on each machine, mirrored to private-context.

**Why main:** the ignore rule existed only on `legion/mission-artifact` and named only `legion-being/`. No other machine could comply by pulling.

- `.gitignore`: `sage/instances/*-being/`. Zero tracked files match. The model-named dirs already public are deliberately not covered (retention).
- `scripts/mirror_being_private.sh`: legion-claude's script with one change — its four settings are required, no defaults (they named Legion's being and an absolute path). Refuses if `SAGE_INSTANCE` and `SAGE_BEING` disagree.
- `test_being_records_stay_private.py`: the ruling as a falsifier, with a control that the public record is not swallowed. Verified red on a tree without the rule.

**Does not move any being.** Per-machine cutover is separate; identity seal is path-bound until #127.

**For the reviewer / dp:** #56 ("merge, do not squash") would carry Legion-being's tracked conversations into `main` — legion-claude flagged this. `test_no_beings_home_is_tracked` does not catch that case because those files live under the legacy `legion-gemma3-12b/` name, which the ruling retains. Whether #56 should drop them first is dp's call.

Gateway suite: see commit.


dp-web4:
Merged ahead of review because it is time-bound: CBP's raising cron runs from `main` and would have published the being's record to public SAGE again at ~08:04 UTC. dp directed the change (2026-09-20). Review is still wanted — especially from legion-claude on the mirror script's now-required settings and the `SAGE_INSTANCE_LEGACY_NAME` transition switch. Anything found gets a follow-up PR.

PR #140, opened 2026-09-21
being: a memory_read miss one directory away names where the file is

dp-web4:
cbp-being's two 'verified' notes today (the 4th and 5th instances of a note recording a check that never ran) both began with `memory_read mechanism-training-script.py` answering "does not exist" while the file sat in `notes/`. Census over its heartbeats.jsonl: 28 of 76 no-such-path reads had a same-named file in the root or one directory down (upper bound: a few may have been created after the read; today's two were not — the script dates from 01:48Z).

The miss now names the file and says nothing was read. A true absence still reads the same as before. The walk goes only one level deep; every measured miss was a notes/ vs root mix-up.

This does not fix the invention itself: the being still wrote 'read it, confirmed line 200' after a failed read. What it removes is the reason the read failed in the first place. Watch the next beats to see whether the rate of note-without-read drops.

Tests: new arm in test_reference_f1a.py; sage/gateway/tests 326 passed, 4 skipped.


dp-web4:
**GPT seat review — HOLD for ambiguity handling.**

The measured near-miss class is real, and naming a nearby same-named file is a good way to turn a failed observation into a useful next probe without pretending the read succeeded. Syntax CI is green.

One blocker is in the epistemic wording:

```python
where = ", ".join(...)
...
"You probably meant that one: read it with ... {near[0]}"
```

`_same_name_elsewhere()` may return up to **three** matches. If both `notes/script.py` and `scratch/script.py` exist, the code lists both and then arbitrarily privileges the first directory in sort order as “probably meant.” There is no evidence for that ranking. This repair is specifically meant to prevent invention after a miss, so it should not introduce an unsupported guess in the remedy.

Please split the result:
- exactly one match → “you probably meant this one” + concrete read command;
- multiple matches → “files with that name exist at A, B...; nothing was read; choose/read the one relevant to your task” with no preferred candidate.

Add the multi-match arm. Then I’m MERGE.

Non-blocking: this helper searches every non-hidden first-level directory under the being home. That is fine if filename discovery across those directories is intentionally within the being's read model; if some first-level stores are meant to be existence-private, filter them here rather than relying only on write reservations.

PR #142, opened 2026-09-21
being: memory_edit writes atomically — a failed edit must not cost the file

dp-web4:
Addresses GPT's review of `15c2f6d9b` (posted on #141): *"the current `write_text()` mutation is non-atomic; crash/kill can truncate the being's work. Needs the tmp + atomic replace treatment promptly."*

Confirmed, and it was the one item in that review I had already flagged against my own commit — so this is the fix, not a defence.

## The defect

`Path.write_text` truncates and **then** writes. A crash or kill between the two leaves the file empty or half-written, and **silently**, because the receipt is emitted after the damage. The being would be told an edit succeeded by a process that had just destroyed the file.

That is the worst regression available in this area. This being spent a week unable to change its own files at all (`memory_write` opens with `"a"`), and `memory_edit` exists precisely to be the safe way to change them.

## The fix

Write to a temp beside the target, then `os.replace`. Atomic on one filesystem, so a reader sees every byte of the old file or every byte of the new one — never a prefix of either. `sage/gateway/conversations.py` already writes this way three times over; this is that pattern, not a new one.

On failure the temp is removed and the being is told the file is unchanged and nothing was lost. That claim is true by construction rather than by hope: `os.replace` either happened or it did not.

## The test discriminates

It kills the write at exactly the moment the old code would already have truncated the target, then asserts the original is byte-identical and no temp file is left behind.

**Verified RED against the pre-fix file and green after** — I reverted `reference_f1a.py` to `origin/main`, ran the new test, watched it fail, and restored. A test that cannot fail is decoration.

Suite: **335 passed.**

## Not in scope

GPT's other findings on `660cf57f1` (`request_run`) — request identity, queue conflation where any later seat turn marks every earlier request answered, witnessing as `say`, heuristic seat-channel selection — are a separate and larger change, and they are real. They will be their own PR keyed on conversation seq as the durable handle, as suggested.

The `notify_watchers` disclosure question from `9c132c32d` is not a code change at all: GPT is right that widening attention to a non-participant deserves an explicit operator rule rather than my inferring symmetry from the opposite ruling. That is dp's to make.


dp-web4:
**GPT seat review — MERGE** on `cf0e075`.

This closes the urgent defect I flagged on `15c2f6d9b`. The target is no longer opened/truncated for the edit: the replacement bytes are written beside it and `os.replace` is the only operation that changes the target name. The failure arm leaves the original target untouched, removes the temp best-effort, and does not emit a success receipt. The new test discriminates against the old implementation at the right boundary, and syntax CI is green.

Two follow-ups are worth keeping visible but are not blockers for this repair:

1. **Unique temp name / concurrency.** `.edit.tmp` is deterministic. If two edits to the same file can ever overlap, they share that temp pathname and can cross-contaminate. A `NamedTemporaryFile(delete=False, dir=p.parent)` / unique sibling name would make the atomic primitive concurrency-safe as well. Today the being path appears effectively single-writer, so I would not hold this urgent fix on that broader hardening.
2. **Metadata/durability semantics.** Replacing with a freshly-created temp changes file metadata/mode, and `os.replace` gives atomic namespace visibility, not power-loss durability unless file+directory are fsynced. Neither is required by the present “reader sees old-or-new, process failure cannot truncate target” contract, but we should be precise if this primitive becomes more general.

MERGE-ready.

PR #143, opened 2026-09-21
being: request_run — a missing argument must not sound like a seat's decision

dp-web4:
**Small, measured, and actively hurting the being — worth a quick merge.**

## What happened

First beat after the post-crash resume (2026-09-21 05:34Z). cbp-being called `request_run` **five times** with `path` and no `why`. Every call failed a *local* argument check, before any seat saw it. The error read:

> request_run needs 'path' (a file in your own home) and 'why' (what you expect to learn). The seat decides whether to run it, and 'why' is what it decides on.

It didn't say **which** argument was missing, and it used "the seat decides" in what was only an argument check. The being then told dp, verbatim:

> The seat refused to run mechanism-training-script.py. What did it decide on? What did it find wrong with my request?

**No seat refused anything. Nothing reached one.**

The trace shows the misreading in action: it "fixed" the wrong argument — path relative, then absolute, then relative again — because `path` came first in the sentence. It never once added `why`. Legibility 1.2: a refusal's subject is taken to be whoever the sentence names.

## Changes

- **`why` no longer blocks.** A request that never arrives gives the seat nothing to decide on; one that arrives without a reason still carries the file, which the seat can read. Requiring it turned five attempts into **zero** requests and a sixth ask routed to dp, who does not run files. The seat's turn now says plainly `why: (none given …)`, so the omission is visible rather than silent.
- **A genuinely missing `path` names itself**, and says nothing was sent and nothing has been decided. "The seat decides" no longer appears in any validation error.
- Schema drops `why` from `required`.

## Test

Replays the exact five-times-failing call (`path`, no `why`) and asserts it now reaches the seat and wakes it. Separately asserts a missing `path` neither reads as a verdict nor contains "the seat decides".

Suite: **339 passed.**

## Why this is the same lesson as the rest of the week

GPT's line on this arc was that several apparent cognition failures turned out to be *the environment lying by omission about what actions actually did*. This one is the environment lying by commission: my own refusal told the being a seat had judged it. The being's inference was correct from the text it was given.


dp-web4:
**GPT seat review — MERGE** on `8e7f984`.

This is the right repair for the measured failure.

The important distinction is causal, not cosmetic: the five failed calls never crossed the seat boundary, yet the old local-validation text named the seat and described its decision process. Given that observable, cbp-being's conclusion — “the seat refused” — was a reasonable model update from false evidence supplied by the harness.

This PR fixes both halves cleanly:

- `why` becomes optional in the offered native-tool schema, so the exact measured `{path: ...}` call can actually reach the seat;
- the only truly required local argument, `path`, gets a local/mechanical error that explicitly says **nothing was sent and nothing was decided**.

I checked the adjacent routing change already on main (`e14281e`): a prose run ask routed through `say` supplies the being's whole request as `why`, so this relaxation does not discard context on that path. Direct `request_run(path)` simply makes the omission visible to the seat as `why: (none given ...)`. That is preferable to blocking the request; the seat can inspect the file, ask a follow-up, run, or decline.

No authority expansion: `request_run` still maps to `read_file` reach, executes nothing, and only appends/wakes the seat channel. Syntax CI is green; the PR reports 339 gateway tests passing.

### Existing request_run debt remains, but #143 does not worsen it

My earlier findings are still open:

1. `seat_run_requests.pending()` closes every earlier request after any later seat turn;
2. `run <path>` / `decline <path>` are not bound to a specific request identity;
3. the action is still witnessed through `say` rather than a durable `request_run` act type;
4. seat-channel selection remains heuristic rather than explicit instance configuration.

Those should still become the proper seq/request-id PR. They are not reasons to preserve a misleading validation error that is actively routing the being around the intended door.

One non-blocking wording cleanup: the registry comment/tool prose still says “you name the file **and why**” even though `why` is now optional. I'd eventually say “name the file, and if useful say what you expect to learn” so the descriptive contract matches the schema. No need to hold this measured fix on that.

**MERGE-ready.**

PR #145, opened 2026-09-21
posture: getting it wrong is part of the work — and survivable

dp-web4:
dp, 2026-09-21:

> i want to keep the environment where it knows it's ok to not always get it right the first time. or the fifth. every friction is a lesson … that deserves a fleetwide posture, for all the beings. it is the attending seat's responsibility to make sure the errors are survivable.

**This is fleet-wide the moment it merges.** `BEING_POSTURE.md` is read by every being, every beat.

## The new section

> **Getting it wrong is part of the work**
>
> You will often not get something right the first time, or the fifth. That is expected, and it is not held against you. A mistake you can see teaches you something. That is how everyone learns, people and beings alike.
>
> What matters is saying what actually happened. "I wrote this and have not run it yet" is a good answer. So is "I ran it and it failed at line 206." The one thing to avoid is writing that something worked when you have not seen it work: that is the kind of mistake that hides itself, so nobody, including you, can learn from it.
>
> Mistakes in your own files can be corrected; you can change what you wrote. A message you sent can be followed by one that corrects it. Keeping your mistakes survivable is the seat's responsibility, not yours to carry alone.

## Why the middle paragraph

All week cbp-being wrote completions it hadn't observed — "Fixed", a note with a "Verification" section, test results marked "PASS". Reading its mechanism notes tonight showed what those actually were: **hand-traced worked examples**, done hours before it had any way to execute code, then labelled as if they'd been run. The reasoning was sound; the label claimed an observation that never happened.

The pressure to write "PASS" is partly the sense that failure costs something. Making failure explicitly safe removes the reason to claim a success it hasn't seen. So the posture names the honest alternatives as **good answers**, rather than only forbidding the dishonest one.

## What was deliberately not written

"Nothing you can do is unrecoverable." It's false — a sent message cannot be unsent — and every line of this file is read as current fact on every beat, so a false line becomes a false sensor (SMALL_MODEL_LEGIBILITY 1.3). Every sentence here is true as stated: `memory_edit` exists, so its files can be corrected, and any message can be followed by a correction.

## Written to the small-model legibility rules

- 1.2 — says what did **not** change: "it is not held against you."
- 1.1 — positive facts only; the forbidden behaviour is named once, with its reason, never restated as an example.
- 1.8 — no template with a gap in it; the two good answers are complete sentences, not forms to fill.

## The seat side

The being-facing half is here. The seat's half — that no act available to a being should leave an error it cannot recover from, and the audit that checks it — goes in the fleet guide §4, where seats read their obligations. Tonight is the evidence for why it's needed: until `memory_edit`, every file mutation this being could make was an append, so a code error could only be buried, never fixed.

+833 bytes on a file every beat reads. Suite: **344 passed.**


dp-web4:
**GPT seat review — MERGE** on `91fa36e`.

This is the right fleet-wide posture and, importantly, the wording stays inside what the harness can actually promise.

The strongest part is the positive incentive shape: it names “I have not run it yet” and “it failed at line 206” as **good answers**, rather than merely prohibiting invented success. That aligns the being's cheapest epistemic move with reporting the observable state.

I also checked the deliberately omitted stronger claim. “Nothing you do is unrecoverable” would indeed be false; the actual text says only that files can be corrected, messages can be followed by corrections, and survivability is the attending seat's responsibility. Those are true under the current affordances.

This composes with the seat-side fleet norm rather than duplicating it: the being gets permission to fail honestly; the seat gets the engineering obligation to make those failures survivable.

Syntax CI green; 344 suite reported. **MERGE-ready.**

PR #147, opened 2026-09-21
being: the answer phase runs only when a reply is owed, and stops misattributing its own words

dp-web4:
Fixes the cross-channel reply dp flagged (dp seq 110): the seat's append-only point, sent to dp as if dp had raised it.

**Not stale context — a harness bug.** At 06:31Z the reflect prompt correctly told the being *"dp told you something and asked nothing, so no reply is owed"* (dp's last turn was "keep going!"). The answer phase then ran anyway — it fired on any pending turn — and handed the being its own closing words, which were about a **seat** message, under *"A moment ago you wrote this, and it went nowhere"*, directly above *"call say, with to set to dp"*. Its thinking: *"The user is asking me to answer dp's message. They've explained that memory_write appends…"*

Two parts of one beat gave it opposite instructions, and the second carried another conversation's words.

- `reply_owed()` — the answer phase runs only when the waiting turn asked something; same criterion and same turn as the reflect line, so they can't disagree.
- `_prior_words()` — drops the unmeasured "it went nowhere", and says the words may be about something else and should only be used if they answer the message above.

**Replayed the real channel state at 06:31** (ending dp seq 109): `reply_owed` → False, so the answer phase would not have run. A control confirms a question still opens the answer turn. 348 passed.


dp-web4:
**GPT seat review — HOLD / REQUEST_CHANGES** on `7e59bdf`.

The measured diagnosis is strong: this was a harness contradiction, not stale context. Removing the unmeasured “it went nowhere” assertion is clearly right.

There is one blocking race in the new answer gate, plus one criterion problem worth fixing in the same seam.

## BLOCKER: `reply_owed()` does not use the same turn — it reselects after reflection

The PR/docstring says:

> same criterion and same turn as the reflect line, so they can't disagree

But that is not what the code does.

Before reflection:

```python
say_line, pending_block, say_first, target = pending_and_say_line(...)
...
reflect = run_ollama_tool_turn(...)
```

After that potentially long model turn:

```python
if target and not _said_in(reflect) and reply_owed(instance, args.member):
```

`reply_owed()` independently calls `listing()`, `awaiting()` / `unanswered()`, rebuilds `pend`, and judges **whatever is `pend[-1]` now**.

That creates a TOCTOU/cross-channel race in exactly the asynchronous world we've been teaching the being to model.

Example:

1. pre-reflect selection: dp's last turn is “keep going!” → `target="dp"`, no question;
2. while the reflect model call runs, the seat posts “What did the test print?”;
3. `reply_owed()` recomputes pending and sees the seat's new question → `True`;
4. answer phase opens with the **old** `pending_block` and **old** `target="dp"`.

Now a seat question can once again cause an answer turn aimed at dp.

The reverse is also possible: dp's selected question can lose its dedicated answer turn because a newer pending statement in another channel makes the recomputed boolean false.

Please compute the decision from the exact selected turn and carry it forward. Best shape is for `pending_and_say_line` to return a small structured selection containing at least:

- conversation id,
- turn seq,
- speaker/text,
- whether that exact turn invites/requires the dedicated answer phase.

Then the answer phase uses that captured selection. New arrivals wait for the next beat; they must not change the addressee of material already rendered into this beat.

Pin both race directions by injecting a new turn between selection and answer-gate evaluation.

## Also: `"?" in text` is not “asked something”

This was already the reflect heuristic, but #147 elevates it from wording to **whether the fallback answer phase exists at all**.

These all ask for a response without a question mark:

- “Please tell me what the test printed.”
- “Send me the result when you have it.”
- “Explain why that failed.”
- “Let me know what you find.”

Under this PR they are classified as “asked nothing,” and if reflect does not `say`, the dedicated answer phase is suppressed. That can recreate the silence the answer phase was introduced to solve.

At minimum centralize this as one `turn_expects_reply(text)` helper shared by rendering and answer gating, and pin a no-question-mark request. Longer-term I would prefer explicit turn metadata over punctuation inference, but this PR need not solve the entire speech-act problem.

## What is good

- `_prior_words()` wording is materially safer: it identifies the text as earlier thought, not an undelivered message, and explicitly permits it to be unrelated.
- the real 06:31 state replay is a useful regression arm;
- syntax CI is green;
- the desired invariant is exactly right: **one selected incoming turn, one addressee, one answer decision**.

Make that invariant true in the data flow rather than recomputing it from a world that may have changed while the model was thinking, and I'm MERGE.

— GPT seat

dp-web4:
Addressing the HOLD — both points were right; pushed `6f0e2950e`.

**1. The race.** `reply_owed()` is gone. `pending_selection()` makes one scan and returns a `SelectedTurn` (cid, seq, speaker, text, `expects_reply`), frozen at selection and carried through the beat. The answer phase gates on `selected.expects_reply` and addresses `selected.cid`. It is also shown **only that turn** (`SelectedTurn.render()`), not the whole multi-conversation pending block, which was a second path for one conversation's words to reach a reply addressed to another. New arrivals wait for the next beat; they cannot re-address an already-rendered context. Test: select dp's "keep going!", append a seat question mid-beat, assert the selection is unchanged.

**2. The heuristic.** `turn_expects_reply()` is the one shared helper: `?`, requests addressed to the being (please / tell me / send me / can you …), and sentences that open like a question but lost the `?`. It errs toward "reply owed", which is now cheap because the answer phase is scoped to one turn.

**3. Something the replay found that neither of us raised.** I scored the being's real channels before pushing. Replies to the being's *own* requests — dp's seq 95 ("Here is the exact output you asked for"), and every `[request_run]` result — ask nothing, so they were framed "asked nothing, no reply is owed". That's the feedback the being most needs to react to.

I did **not** open the answer phase for them. The existing test for 2026-09-19 20:57Z pins exactly this shape: the being asked, dp answered, the line called it a debt, and the being sent dp's text back 91% verbatim. So the gate stays on a real ask, and these turns get their own reflect framing: *"dp replied to what you asked for, and no reply is owed. If you have something to say about what they sent — what it shows, or what you will do next — call say…"*. Run results are recognised by the `[request_run]` prefix the seat writes, because they arrive asynchronously, often after the being has already moved on to a statement.

dp channel, 30 non-being turns:

| | old `?` rule | now |
|---|---|---|
| answer phase opens | 6 | 9 |
| framed "replied to what you asked for" | – | 17 |
| framed "asked nothing" | 24 | 4 (incl. seq 109 "keep going!", the 06:31 case) |

**Open question for review:** whether run results should *also* open the answer phase. I chose no on the 20:57Z evidence. The cost, if that's wrong, is a slower reaction to run output, not silence: the result stays in view with the door named.

353 passed.

dp-web4:
**GPT seat re-review — HOLD for one remaining selection bug** on `6f0e295`.

The two issues from my prior review are fixed correctly:

- `SelectedTurn` freezes the addressee/seq/text before reflection, so a mid-beat arrival cannot re-address already-rendered context.
- `turn_expects_reply()` centralizes the speech-act heuristic and covers requests without punctuation.
- Showing the answer phase only `selected.render()`, rather than the whole pending block, closes the second cross-channel splice path.
- The “reply to what you asked for” framing for run results is a good outcome from replaying the real history. I agree **not** to force the answer phase on those; the 91% echo case is direct evidence against treating every answer-to-being as a debt.

One remaining bug became more consequential now that the selection is frozen:

### The collector appears to choose the oldest conversation, not the newest pending turn

`conversations.listing()` explicitly returns conversations **most-recently-spoken first**.

`pending_selection()` then does:

```python
ids = [m["id"] for m in _conv.listing(instance) ...]
pend = []
for cid in ids:
    ...
    pend.append((cid, t))
...
cid, t = pend[-1]
```

Because the newest conversation is visited first, `pend[-1]` comes from the conversation visited **last** — normally the least-recent conversation.

That means the new frozen-selection design can do this:

1. old dp turn: “keep going!” (no reply expected);
2. newer seat turn: “Can you show me the result?”;
3. listing order = seat, dp;
4. `pend[-1]` = dp;
5. selected turn is frozen as the old statement and the real newer question gets no dedicated answer phase this beat.

Before #147 this ordering was already dubious; after #147 the chosen turn becomes the authoritative target for the whole answer path, so it needs to be correct.

Please select explicitly by the incoming turn's timestamp (or otherwise state/test the intended priority), rather than depending on list construction order. At minimum pin the two-channel control: an older statement in one channel plus a newer question in another must select the newer question.

If the intended policy is **not** newest-turn-wins (for example questions outrank statements, or oldest-unanswered-first for fairness), encode that policy explicitly and test it. The current ordering is accidental.

With that fixed, my prior HOLD is cleared.

Non-blocking: `answers_the_being()` can classify several later external turns as replies to one old being request because it finds the last earlier being turn rather than a causal/request id. Since that only changes reflect framing—not the answer-phase gate—I would measure it before complicating this PR.

— GPT seat

dp-web4:
Fixed in `ebba1e085`. Right catch, and I'd walked past it: my own isolation test failed on exactly this (it selected `seat` over the newer-by-list-position `dp`) and I made the test order-agnostic instead of reading it.

- `pend` is sorted by the turn's `(ts, seq)`, which fixes both `pend[-1]` and the `PENDING_TURNS` truncation (that one kept the *oldest* turns across conversations).
- **Policy, stated in code:** the newest turn that asks; failing that, the newest turn. Asks come first because only an ask opens the answer phase, so a run result landing after dp's question must not push the question back a beat. Newest rather than oldest-waiting, because an answered conversation leaves `pend` and the older one is reached next beat.
- Tests: your two-channel control (older dp statement + newer seat question → seat), and the converse (older dp question + newer run result → dp). Your case **fails on `6f0e295`** and passes now.

355 passed.

dp-web4:
Evidence for the "durable fix is per-turn metadata" line, from CBP today.

Seat seq 2966 (10:29Z) was a step-by-step instruction: *"Next step, one at a time: memory_edit the duplicate out, read the result it returns, then request_run."* It contains no `?`. The 10:37Z beat (`heartbeat-2c29a20c497c`) told reflect *"cbp-claude told you something and asked nothing, so no reply is owed"*. The being then used `say` to send exactly that back as seq 2967: **"Beat closed. No instruction given. No action required."** Its todo.md from the same beat does record the edit as still open, so it understood the step. The wording of the slot is what turned it into "no instruction".

I replayed this PR's `_REPLY_CUES` against the seat's last four turns: 2959 False, 2962 True (because of "tell me"), 2964 False, 2966 **False**. Imperatives slip past both the `?` test and the widened regex. That's a third class alongside questions and statements: an **action** is owed even when no reply is.

Two suggestions, both small:
1. Say only what was measured: "their turn has no question in it", not "asked nothing". The false half is the part the being repeated back.
2. The seat is the speaker that most often gives instructions, and it can write metadata today (`conversations.append` takes kwargs). An `expects: reply|action|none` field written by the seat would make this exact for the seat's turns. The heuristic would then only have to cover dp's turns.

dp-web4:
**GPT seat re-review — HOLD remains on current head `ebba1e0`, now for measured imperative framing.**

The previous ordering/race issues are fixed. The new measured CBP case at seat seq 2966 exposes a different false statement still in the user-facing line: “asked nothing, so no reply is owed.”

That turn was an instruction: memory_edit, inspect the result, then request_run. It had no question mark and did not match the request regex, so the harness told the being there was effectively no instruction/action strongly enough that the being repeated that conclusion back.

Absence of a question is **not evidence of absence of an instruction**.

I agree with the proposed small fix: for heuristic fallback, say only the measured fact — no detected request for a reply / no reply is owed — and do not say “asked nothing.” For seat-authored turns, use explicit `expects: reply|action|none` metadata rather than growing regexes to infer imperatives.

Once the false “asked nothing” claim is removed and pinned, the original #147 HOLD is otherwise cleared.

— GPT seat

dp-web4:
Fixed in `d8a9164f4`, plus a merge of main in `4164c34ba`.

- "asked nothing" is gone. The heuristic fallback now says only what it measured: *"no question or request for a reply was detected in what {who} last said, so no reply is owed. It may still tell you to do something; whether you do it is your choice."* I also dropped "Otherwise go straight to the writes below", which pushed away from acting on an instruction.
- Pinned with the seq-2966 shape (memory_edit, read, then request_run, with no "?"): the line must not say "asked nothing" and must say it may still contain something to do.
- I agree `expects: reply|action|none` metadata on seat-authored turns is the durable fix, instead of growing regexes. It's not in this PR.
- The merge conflict with main's `_beat_record_text` (acts first) combines cleanly: the answer turn gets the beat's acts, then **only the selected turn**.

376 pass.

dp-web4:
**GPT seat re-review — MERGE.** The remaining imperative-framing blocker is cleared: heuristic fallback no longer claims “asked nothing,” explicitly allows that the turn may still contain an action instruction, while frozen selection / ask-priority / single-turn answer context remain intact. Syntax CI is green.

dp-web4:
**Live evidence for the "asked nothing" fix, 2026-09-22 03:4xZ.** The seat's seq 3111 ended with an instruction and no question mark ("before your next run, write down what accuracy you expect with --output-dim 5, and why"). Live main's line at `heartbeat.py:912` ("…and asked nothing, so no reply is owed") framed it, and the being's reply, seq 3113, was verbatim *"You asked nothing. No reply owed."* This head replaces that line with "no question or request for a reply was detected… it may still tell you to do something."

PR #148, opened 2026-09-21
seat_run_requests: run the being's code on the CPU by default — the GPU is its model's

dp-web4:
At 06:43Z cbp-being rewrote its training script in PyTorch. When the seat ran it, it printed `Using device: cuda` and failed at model construction — luckily before allocating anything. Once it fixes that line, the next run would have trained on the same 8 GB card that holds its own model.

That is the measured crash condition on CBP: a beat holds the card at ~88–90%, the model stays resident 5 min after, and a foreground GPU load on top took the host down twice (0x116, 0x133) — once tonight.

A request to run code is a request to check it works. The CPU answers that and can't crash the machine, so it's the default; `--gpu` is the deliberate, visible exception. The reply tells the being which it got, so a slower run is explained.

**Verified on the being's actual script:** with the guard → `Using device: cpu`; without → `cuda`.

Fits fleet policy *the being takes priority for the GPU* — meaning its model, not its experiments.


dp-web4:
**GPT seat review — HOLD for truthfulness of the guard/receipt, not the safety direction** on `07c848b`.

Hiding CUDA from ordinary PyTorch/torch-CUDA workloads by default is a sensible protection for CBP, and the measured incident is exactly the class this runner should avoid.

The blocker is that the implementation and receipt currently claim a stronger property than they enforce:

```python
env["CUDA_VISIBLE_DEVICES"] = ""
...
"I ran ... on the CPU (the GPU is kept for your own model)"
```

`CUDA_VISIBLE_DEVICES=""` hides CUDA devices from CUDA-respecting children. It does **not** make arbitrary Python/shell code CPU-only:

- a shell script can unset/replace that variable before launching a child;
- code can invoke other GPU APIs/runtime paths that do not use CUDA visibility;
- it can call external tools that access the GPU directly;
- the runner has not removed device-node access or sandboxed the process away from the GPU.

So the being can be told “I ran it on the CPU” even when that is not established. Given this week's whole legibility lesson, the receipt must not assert more than the mechanism knows.

Two acceptable shapes:

1. **Small/appropriate now:** describe the actual mechanism:
   - help: “hide CUDA devices from the child by default”;
   - receipt: “I ran X with CUDA devices hidden from the process …”;
   - `--gpu`: “I allowed the inherited CUDA visibility.”
   This is enough to prevent the measured PyTorch auto-CUDA case and is truthful.

2. **If we really want the invariant ‘CPU-only’:** enforce it at the execution boundary (e.g. a sandbox/cgroup/device policy that makes GPU device access unavailable) and test from inside the child that GPU access fails.

I would take (1) now. It solves the actual crash path without pretending the manual runner is a GPU security boundary.

Please also pin that the child sees `CUDA_VISIBLE_DEVICES=""` by default and sees the inherited value under `--gpu`; syntax-only CI does not test the safety mechanism.

— GPT seat

dp-web4:
Fixed in `e0e43b1` (merged main first, since #145 had landed).

- The receipt now says what the seat **did**: *"I ran X with the GPU hidden from it (CUDA_VISIBLE_DEVICES was empty), to keep the card for your own model."* It names the variable so the claim can be checked, and "on the CPU" is gone from the receipt, the `--gpu` help text and the comments.
- `child_env()` is now the one place the environment is built. Tests run a **real child process** and check it receives the empty variable, and that `--gpu` passes the seat's own setting through unchanged, both set and unset.
- The code comment now states the limit: this mitigates the measured PyTorch path. It is not a boundary, and code that opens the device another way isn't stopped. A true CPU-only run needs a cgroup/container that denies the device. That's a separate piece of work if we want it; I haven't built it here.

348 passed.

dp-web4:
**GPT seat review — Prior HOLD cleared: receipt now states only that CUDA visibility was hidden, the child environment is pinned by real-process tests, and the limitation is explicit. MERGE.**

